Issue #413, opened 2014-09-19
Remind users of different types of conservation equations


fipymigrate:
Users may find the behavior of the coefficient to the various terms surprising if they're unfamiliar with looking at different types of conservation equations. This ticket is to add a simple example demonstrating the difference in handling non-uniform mass diffusivity and non-uniform thermal diffusivity within the mesh1D diffusion example.

 _Imported from trac ticket [#670](http://www.ctcms.nist.gov/cgi-bin/redirectLegacyMatforge.py?url=http://matforge.org/fipy/ticket/670),  created by raybsmith on 08-18-2014 at 15:28, last modified: 08-18-2014 at 15:28_

guyer:
Fixed by #423

PR #423, opened 2014-09-25
Fix ticket:413 by adding section to mesh1D example

raybsmith:
Add a section to the `mesh1D` diffusion example showing the difference
between spatial parameter variation in heat transfer as compared to
mass transfer. Also, add relevant figures to the `generated` folder.


guyer:
Addresses #413

PR #439, opened 2015-04-07
Corrected two tests

alfrenardi:
Two trivial issues with the test have been corrected.


wd15:
Thanks for the fixes. Much appreciated.

guyer:
These changes aren't really robust. They're appropriate for recent releases of NumPy, but the older tests used to be right.

alfrenardi:
That may be true. Is there a standard method (at least regarding the convections used in FiPy) to deal with changes between different versions of the same library?

guyer:
Not really. If you read through fipy/tools/numerix.py, you'll find a lot of special case code to account for how the libraries changed along the way.

PR #441, opened 2015-07-29
compare gmsh versions as tuples, not floats

tkphd:
Parallel FiPy hangs on test case (https://gist.github.com/anonymous/43afb6c567ebfd6d3b2f) after rejecting Gmsh 2.10.0, since float(2.10) < float(2.5). Comparing versions as '.'-delimited tuples fixes the error.


guyer:
I think we should do what's in #442 instead.

PR #442, opened 2015-08-01
Robust Gmsh version checking with distutils.version.StrictVersion

guyer:
`distutils` in the Python standard library provides robust version checking, which is superior to the proposal in #441 and vastly superior to what FiPy was doing before.


guyer:
Botched the comment

guyer:
Never mind. `git push --force` fixed the comment.

tkphd:
Tested: result is equivalent to #441, but #442 is more robust and likely future-proof.

Issue #455, opened 2015-08-25
setup.py gist warning


tkphd:
If gist is not installed, executing python setup.py install warns, "The optional module(s) gist cannot be loaded. FiPy will have improved capabilities if these modules are installed." After merging #453, gist is no longer a FiPy dependency, and this warning is in error.

guyer:
Fixed by #456

PR #456, opened 2015-08-25
purge gist from viewers and optional module lists in setup.py

tkphd:


guyer:
should purge gnuplot, too

tkphd:
Done.

Addresses #455

PR #460, opened 2015-09-28
Fix typeset analytical solution

guyer:
Typset equation for C did not agree with `Canalytical` (nor did it satisfy the boundary conditions).


tkphd:
Typesetting properly builds as PDF and agrees with `Canalytical` expression. Merged changes and deleted branch.

Issue #378, opened 2014-09-19
coupled Cahn-Hilliard example doesn't run properly


guyer:
Although [examples/cahnHilliard/mesh2DCoupled.py@fe00adaf9e](../tree/fe00adaf9e/examples/cahnHilliard/mesh2DCoupled.py) passes tests, any attempt to increase the size of the mesh causes problems:
- At 100<sup>2</sup>, PySparse needs over 350 MB, likely due to `LinearLUSolver`.
- Trilinos does not give a sensible solution to the coupled problem, but does fine with the vector equation. They seem to be equivalent.

Although the tests, as written, pass (not that we check very much), we should document that PySparse is not suitable and we need to figure out what's wrong with the coupled treatment.

 _Imported from trac ticket [#582](http://www.ctcms.nist.gov/cgi-bin/redirectLegacyMatforge.py?url=http://matforge.org/fipy/ticket/582),  created by guyer on 03-26-2013 at 13:45, last modified: 09-06-2013 at 10:04_

guyer:
The changes in 8a1f81d 1e56dc5 seem correct, but PySparse still uses a deranged amount of memory and Trilinos doesn't seem to evolve at all (with the default solver)

guyer:
[PETSc]() also has problems. Conversely to Trilinos, PETSc can solve the coupled form, but not the vector form. The difference appears to be a bunch of non-zero zeros entered in the matrix for the vector form, e.g., 
# vector
```
0 convergence: KSP_CONVERGED_RTOL 
1 convergence: KSP_CONVERGED_RTOL 
0 iterations: 5 / 1000 
1 iterations: 5 / 1000 
Mat Object: 2 MPI processes
  type: mpiaij
row 0: (0, 9.27582)  (1, -0.)  (2, -0.)  (4, 2.)  (5, -1.)  (6, -1.) 
row 1: (0, -0.)  (1, 9.27582)  (3, -0.)  (4, -1.)  (5, 2.)  (7, -1.) 
row 2: (0, -0.)  (2, 9.27582)  (3, -0.)  (4, -1.)  (6, 2.)  (7, -1.) 
row 3: (1, -0.)  (2, -0.)  (3, 9.27582)  (5, -1.)  (6, -1.)  (7, 2.) 
row 4: (0, -2.)  (1, 1.)  (2, 1.)  (4, 0.0625)  (5, -0.)  (6, -0.) 
[...]

PR #463, opened 2015-10-01
Correct implementation of examples.cahnHilliard.mesh2DCoupled. Fixes …

tkphd:
…#378.

The numerical implementation introduced a stray - sign in the linearized
source term in $\psi$.  Corrected expression allows simulations of
reasonable duration, and shows desired spinodal decomposition from random
initial condition.

### DIFF
--- a/examples/cahnHilliard/mesh2DCoupled.py
+++ b/examples/cahnHilliard/mesh2DCoupled.py
@@ -86,7 +86,7 @@ We start the problem with random fluctuations about :math:`\phi = 1/2`
 :term:`FiPy` doesn't plot or output anything unless you tell it to:
 
 >>> if __name__ == "__main__":
-...     viewer = Viewer(vars=(phi, psi)) # , datamin=0., datamax=1.)
+...     viewer = Viewer(vars=(phi, psi), datamin=0., datamax=1.)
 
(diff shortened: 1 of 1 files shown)


tkphd:
After Taylor series expansion of the source term about ```$\phi=\phi_n$```,

```math
\psi &= \frac{\partial f}{\partial\phi_n} 
        + \frac{\partial^2 f}{\partial\phi_n^2}\left(\phi - \phi_n\right)
        - \epsilon^2\nabla^2\phi\\
     &= \left(\frac{\partial f}{\partial\phi} 
              - \frac{\partial^2 f}{\partial\phi^2}\phi\right)_{\mathrm{old}}
        + \frac{\partial^2 f}{\partial\phi^2}\phi.
        - \epsilon^2\nabla^2\phi
```

The numerical implementation introduced a stray (-) sign, therefore representing

```math 
\psi &= \left(\frac{\partial f}{\partial\phi} 
              - \frac{\partial^2 f}{\partial\phi^2}\phi\right)_{\mathrm{old}}
        - \frac{\partial^2 f}{\partial\phi^2}\phi.
        - \epsilon^2\nabla^2\phi
```

guyer:
Fixes #378

Issue #315, opened 2014-09-19
move away from from fipy import *


wd15:
After my experiences with python-meep, I think the convenience of using \* is not worth the trouble it can cause when one doesn't know where stuff really comes from. Using

```
from fipy import CellVariable, Grid2D, Viewer, TransientTerm, DiffusionTerm
```

is not really a big deal IMO.

 _Imported from trac ticket [#437](http://www.ctcms.nist.gov/cgi-bin/redirectLegacyMatforge.py?url=http://matforge.org/fipy/ticket/437),  created by wd15 on 04-06-2012 at 17:22, last modified: 09-06-2013 at 07:27_

PR #466, opened 2015-10-05
Update import syntax in examples

tkphd:
Explicitly import FiPy modules as needed, instead of everything.

Addresses #315


tkphd:
FYI, each example was executed stand-alone from the command line, and through each directory's  test.py. Except the levelSet/electrochem examples which depend on custom Mayavi viewers, the updated examples cleanly exited in both execution modes. The electrochem examples would benefit from expert attention -- bug report to follow.

guyer:
Addresses #315. Still much `from fipy import *` in the modules.

Issue #467, opened 2015-10-07
Out-of-sequence Viewer imports


tkphd:
The generic Viewer constructor (defined in ```fipy/viewers/__init__.py```) works as advertised -- it imports the first viewer it discovers that can handle the variable's dimensionality -- but perhaps not as expected, i.e. Matplotlib before Mayavi.

In setup.py, the entry_points are ordered (matplotlib, mayavi), and the FiPy documentation implies that MatplotlibViewer should be called unless the problem is 3-D. However, the Viewer() constructor uses ```pkg_resources.iter_entry_points```, which places no guarantees on the order in which entry points are yielded. Its documentation (https://pythonhosted.org/setuptools/pkg_resources.html#entry-points) suggests entry points are yielded in the order they are found on ```sys.path```, which is environment- and system-dependent.

This issue cropped up (and is reproducible) on two Linux machines with FiPy installed in conda environments using Python 2.7. If both matplotlib (1.4.3) and mayavi (4.4.0) are installed using the conda package manager, the mayavi entry point is yielded first:
```Python
>>> from fipy import Viewer
>>> from pkg_resources import working_set
>>> for ep in working_set.iter_entry_points(group='fipy.viewers',name=None):
>>>    print ep
mayavi = fipy.viewers.mayaviViewer:MayaviClient
matplotlib = fipy.viewers.matplotlibViewer:MatplotlibViewer
```
In this order, even 1-D examples open Mayavi to display results. This occurs despite the correct sequence in setup.py and the entry_points.txt it generates:
```bash
$ cat FiPy.egg-info/entry_points.txt 

            [fipy.viewers]
            matplotlib = fipy.viewers.matplotlibViewer:MatplotlibViewer
            mayavi = fipy.viewers.mayaviViewer:MayaviClient
```

On these machines, removing mayavi from the conda environment and re-installing it through the operating system's package manager (i.e., dpkg) resolves the problem, but in the clumsiest possible way.

tkphd:
Strangely, after purging both Mayavi and Matplotlib from the machine and re-installing through the conda package manager, neither are explicitly listed in the ```working_set```:
```Python
>>> print pkg_resources.working_set.entries
[
    '',
    '$ANACONDA/lib/python2.7/site-packages/Cython-0.22.1.egg',
    '$ANACONDA/lib/python2.7/site-packages/scikit_fmm-0.0.6.egg', 
[...]

PR #469, opened 2015-10-07
Sort entry points by package name before testing.

tkphd:
Addresses #467 as an example fix. A more robust method would be preferable.


guyer:
A better idiom is:
```python
    for name, ep in sorted(enpts):
        attempts.append(ep.name)
```

tkphd:
Makes sense.

Issue #432, opened 2014-11-03
Python 3 issues


wd15:
Issues raised by Stefan on the [mailing list](http://thread.gmane.org/gmane.comp.python.fipy/3631/focus=3657). 
- [x] TabError in Python 3 of the kind

```
  Sorry: TabError: inconsistent use of tabs and spaces in indentation
>> (testBase.py, line 47)
>>   Sorry: TabError: inconsistent use of tabs and spaces in indentation
>> (trilinosSolver.py, line 148)
>>   Sorry: TabError: inconsistent use of tabs and spaces in indentation
>> (gistVectorViewer.py, line 70)
```
- [x] Automated message saying that PySparse is required in Python 3 when in fact it isn't

```
>> After fixing the files, I got that far:
>>
>>   The required module(s) ['pysparse'] cannot be loaded.
>>   FiPy will not work properly until these modules are installed.
>>   ------
>>   The optional module(s) ['gist', 'mayavi'] cannot be loaded.
>>   FiPy will have improved capabilities if these modules are installed.
>>
>> It seems `pysparse` is only available for Python 2. Is this
>> correct? Is it indeed needed?
>
> You don't need PySparse. It isn't available for Python 3 yet. The Scipy
> solvers are adequate to get started. None of the solvers are required
> packages, but you obviously need one of them.

I suggest fixing the message then because it says `pysparse`
is required. :-)
```
- [ ] move the Python 3 support documentation to INSTALLATION.txt
- [x] make FiPy pip installable for Python 3
- [ ] Use the 2to3 version of the code as default so that users employ 3to2 if they want to use 2.6 or 2.5 versions.

guyer:
Resolved by #645

PR #473, opened 2015-11-08
Single source Python 2 and 3

pya:
Many new Python users (rightfully) start with Python 3. Therefore, having FiPy support Python 3 out-of-the box would be very valuable. 

This version runs all 195 tests in `example/test.py` with Python 2.7, 3.4, and 3.5 (with the same source). It uses the very good library `future` (http://python-future.org/). This allows to test all changes immediately against Python 2.7 **and** Python 3. 

Further development can take advantage of `future` by programming in Python 3 style (think `range`, `zip`, `map` etc.) and still support Python 2.7. In the distance future, when Python 2.7 support is discontinued, only the imports from `future` need to be removed to get a pure Python 3 version.


guyer:
First off, thank you *very much* for this contribution!

There are a few issues I'd like to see addressed before adoption:
- [ ] numpy 1.10 `axis=` has nothing to do with 2 to 3 migration and clobbers #475 & #477 that I entered before seeing this pull request.
- [ ] Changes to trailing whitespace shouldn't be part of this pull request (unless required by Py3k, which I'm unaware of). I'm not opposed to these changes (although I wish git would stop making such a fuss about something that's none of its business), but they presently clutter up this pull request to the point that it's very hard to see the "real" changes.
- [ ] `old_div` strikes me as ugly. I think we have a clear enough idea of when floating and when integer division is called for that we should be selectively using `/` or `//`. `old_div` just masks the problem of ambiguous division with a syntax that's hard to read.
- [ ] In general, there's too much going on in each commit for me to effectively evaluate, particularly when I approve of most of it, but not all. If the `futurize` command (assuming that's what you used) has any options to control what it converts, I'd like to see individual, atomic commits that each address a single issue: `1` -> `True`/`0` -> `False` in one, `raw_input` -> `input` in another, `print` in another, and so on.
- [ ] I'm inherently pretty conservative about making backward-incompatible changes and [python-future](http://python-future.org) will require a big change in the minimum version of Python. I've posted a [survey](https://www.surveymonkey.com/collect/?collector_id=77356212) to the [mailing list](http://permalink.gmane.org/gmane.comp.python.fipy/3919) to see if this will inconvenience very many of our users (I'm guessing not, but I'd like to be sure).

Addresses #432

pya:
Thanks for your response. All your points are totally valid. Let's see this PR as a proof-of-concept that has some (serious) "procedural" issues.

* Sorry for mixing up the NumPy 1.10 issue here. I did not pay enough attention.
* I was not aware that there is no option  to suppress white space changes in diffs in PRs. This is just a artifact of my editor being setup like this.
* The `old_div` was not intentional but a result of the use of `futurize`. See next bullet point.
* `futurize` offers 52 different kind of fixes. We might not want/need all of these (see previous bullet point). Applying one fix at a time and commit each one separately sounds like a good idea.
* Python 2.4 seems really old. The only place I need to work with Python 2.5 is in embedded environment where I have no way to determine what Python version I can use. I am really happy with `conda` environments that allow a clean separation between different combinations of Python and library versions on the same machine.  Will see what the survey results say.

I would suggest to take this PR as a basis to work from. Basically, don't merged anything into FiPy now but re-do things based on what I've learned. Next steps cold be:

1. Wait for the survey results. If the minimum Python version moves up to 2.6, we can proceed, otherwise it will be very difficult to achieve out-of-the-box Python 3 support. Is there a deadline for the survey? If not, setting one might speed up answers.
2. If Python 2.6 becomes the minimum required version we can proceed, addressing the issues above, i.e. after taking out the NumPy version change and the trailing newlines deletes, go one fix at a time  and commit each one separately. Simultaneous Python 2.4 and Python 3 support would be rather challenging.  
3. It would be great to have more feedback from you and other core developers along the way. Maybe breaking things down into multiple PRs, even though they would not work with Python 3 yet but all tests pass with Python 2, could be a solution. I am open to suggestions such as some kind of branching scheme or whatever works best according to your experience.

guyer:
OK, great. I think we're on the same page. 
- If the survey shows a substantial base still tied to Python 2.4 or 2.5, it should still an option to use `six`. For the record, I don't think that's what the survey will show. Good idea about setting a deadline.
- I'm happy to give feedback along the way. We can either do it via incremental pull requests, or I can just subscribe to your repo and make comments as you push to github.

guyer:
The [survey results](https://www.surveymonkey.com/analyze/6jME9w507Ma8jLeR7TrHqNVGdRMbX5HDVrmgZVLb0u4_3D) are unambiguous (although I'd like to think we have more than 29 people using FiPy). Please proceed with using python-future to make FiPy work with both Python 2.x and 3.x. Let me know what branch you're working on and I can monitor and comment there.

pya:
Great to hear. I will get started. BTW, I cannot see the survey results. I only get a survey monkey login page.

guyer:
Survey results are now published at https://www.surveymonkey.com/results/SM-2CDMMVQJ/

woodscn:
Any updates on this?

wd15:
Sorry @woodscn, neither @wd15 nor @guyer have had time to address this, but I hope to do a round of Python development within the next 6 months and this pull request will be top of the agenda unless @guyer doesn't merge it first.

woodscn:
Fair enough. I'll probably end up maintaining a straight "futurize" branch for the time being, then. At least it sounds like backward compatibility is only necessary as far back as 2.6, which makes things much easier.

guyer:
Actually, there's been nothing to address. @pya was going to submit incremental pull requests. Somebody else is welcome to take this on, but this pull request as written will never be merged.

pya:
Thanks for bringing this up again. I was about to start working on this but had a few urgent things on my plate for the last months. I will probably do something here,  starting in a week or so.

woodscn:
I've been working on it too, in the meantime. I broke up the process of futurizing the code into a few smaller steps, and I've been making sure that tests pass (in Python 2) as I go. You can check out my branch here: https://github.com/woodscn/fipy

guyer:
We completely understand, Mike. Glad to hear you both are willing to tackle this.

achennu:
I am enjoying using this library a lot, only tempered by having to go back to python2. I would be able to find some effort in the coming weeks to help the python 3 port.

Can somebody please explain what's the best place to contribute this effort?

guyer:
@achennu  Thanks for offering, but I'm not sure what you mean by "best place to contribute".  Do you mean how do you do a pull request? Or how do you do the Py3k conversion? Or ...?

Ultimately, we need a pull request against usnistgov/fipy. That pull request can be forked from @woodscn's effort, or you can start fresh. The items I requested of @pya in this pull request would all still apply.

achennu:
Yes I know how to make PRs. I have ran a 2to3 version before and it mostly
ran. While trying to solve equations though, I got errors relating to wrong
opcodes. As I didn't know how to solve that, I stopped.

I can look into this. Just to clarify: idea would be to move to a common
code supporting python 2.7+ and 3.5+.

guyer:
As illustrated [here](https://travis-ci.org/usnistgov/fipy/jobs/359546741), `2to3`conversion does work for `develop` (usnistgov/fipy@43bbbd65ea1867fe73b0a58e00834aec5a616dc4).

I recommend you use [futurize](http://python-future.org/automatic_conversion.html) and preferably change one thing in fipy at a time, i.e., I would prefer one commit per fixer.

guyer:
Fixed by #645

PR #477, opened 2015-11-25
Remove incorrect `axis` argument to `concatenate`

guyer:
This never should have worked, but numpy 1.10 actually checks.
Also provide some documentation for what `_faceToCellDistanceRatio` is and
why it's scalar.

Addresses #475

Issue #475, opened 2015-11-24
Grid3D broken by numpy 1.10


guyer:
As of numpy 1.10, any uses of `Grid3D` fail, e.g.
```python
======================================================================
FAIL: _test (fipy.meshes.uniformGrid3D.UniformGrid3D)
Doctest: fipy.meshes.uniformGrid3D.UniformGrid3D._test
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/guyer/anaconda/envs/py2_7_debug/lib/python2.7/doctest.py", line 2226, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for fipy.meshes.uniformGrid3D.UniformGrid3D._test
  File "/Users/guyer/Documents/research/FiPy/fipy/fipy/meshes/uniformGrid3D.py", line 541, in _test

----------------------------------------------------------------------
File "/Users/guyer/Documents/research/FiPy/fipy/fipy/meshes/uniformGrid3D.py", line 667, in fipy.meshes.uniformGrid3D.UniformGrid3D._test
Failed example:
    print numerix.allclose(faceToCellDistanceRatios, mesh._faceToCellDistanceRatio, atol = 1e-10, rtol = 1e-10) # doctest: +PROCESSOR_0
Exception raised:
    Traceback (most recent call last):
      File "/Users/guyer/anaconda/envs/py2_7_debug/lib/python2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "<doctest fipy.meshes.uniformGrid3D.UniformGrid3D._test[49]>", line 1, in <module>
        print numerix.allclose(faceToCellDistanceRatios, mesh._faceToCellDistanceRatio, atol = 1e-10, rtol = 1e-10) # doctest: +PROCESSOR_0
      File "/Users/guyer/Documents/research/FiPy/fipy/fipy/meshes/uniformGrid3D.py", line 278, in _faceToCellDistanceRatio
        numerix.ravel(YZdis.swapaxes(0,2))), axis=1)
    IndexError: axis 1 out of bounds [0, 1)
```

The issue is that `numerix.ravel` returns a 1D `array` and so `axis=1` is nonsensical in `_faceToCellDistanceRatio`. It looks like this was always wrong (going back to [at least 2011](https://github.com/usnistgov/fipy/blame/develop/fipy/meshes/uniformGrid3D.py#L276)), but numpy tolerated it until now. 

[...]

Issue #470, opened 2015-10-20
Mesh3D cellToCellIDs is broken


pkgw:
Unless I'm strongly confused, the implementation of `_cellToCellIDs` for `UniformMesh3D` is broken, as demonstrated by a simple example:

```python
>>> import fipy
>>> t = fipy.Grid3D (nx=2, ny=2, nz=2)
>>> print (t._cellToCellIDs)
<<< [[-- 0 -- 2 -- 4 -- 6]
<<<  [1 -- 3 -- 5 -- 7 --]
<<<  [-- -- -2 -1 -- -- 2 3]
<<<  [4 5 -- -- 8 9 -- --]
<<<  [-- -- -- -- 0 1 2 3]
<<<  [4 5 6 7 -- -- -- --]]
```

Note that the middle of the array contains both negative numbers and an 8 and a 9, while the maximum legal index is 7. This codepath gets triggered, leading to an exception, if you try to use an AdvectionTerm on a 3D mesh.

[The function](https://github.com/usnistgov/fipy/blob/develop/fipy/meshes/uniformGrid3D.py#L169) is short but it seems to be implement in an (overly?) tricky way so it's not obvious to me how to fix it.

pkgw:
After some total cargo-cult coding, [this patch](https://gist.github.com/pkgw/59dceaa4eba79f513b41) seems to be what I need to get things to run without crashing. I don't understand if they actually work. 

I'm not making a proper pull request since I'm not sure if these fixes are correct, and test cases should be developed if things were truly broken. It certainly seems to me that the current 3D grid implementation causes crashes, though.

guyer:
Your fix looks right to me. A pull request would be most welcome. Tests that demonstrate the problem would be gravy.

PR #478, opened 2015-12-09
fipy/meshes/uniformGrid3D.py: fix _cellToCellIDs and more concatenate() calls

pkgw:
As far as I can tell, there were several codepaths in the UniformGrid3D class that seemed not to have been tested since they didn't work at all; in particular, the _cellToCellIDs map was broken. This commit fixes that (I hope) and a couple of more invalid `axis=n` keywords to `numerix.concatenate()`. Unfortunately I have not put together test cases for these problems, so I'm not completely sure that the "fixed" _cellToCellIDs is always correct.

See issue #470, issue #475, and issue #477.


guyer:
Added tests and merged to develop in 33d033d

Issue #476, opened 2015-11-24
setuptools 18.4 breaks test suite


guyer:
As of setuptools 18.4 (probably 18.0), running the FiPy test suite raises an exception:
```python
$ python setup.py test
/Users/guyer/anaconda/envs/py2_7_debug/lib/python2.7/site-packages/setuptools-18.4-py2.7.egg/setuptools/dist.py:294: UserWarning: The version specified (u'3.1-dev74-gb542e0c') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
running test
running egg_info
writing FiPy.egg-info/PKG-INFO
writing top-level names to FiPy.egg-info/top_level.txt
writing dependency_links to FiPy.egg-info/dependency_links.txt
writing entry points to FiPy.egg-info/entry_points.txt
reading manifest file 'FiPy.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no previously-included files matching '*.png' found under directory 'examples'
warning: no previously-included files matching '*.pdf' found under directory 'examples'
warning: no previously-included files matching '*.png' found under directory 'fipy'
warning: no previously-included files matching '*.pdf' found under directory 'fipy'
warning: no files found matching '*.msh' under directory 'fipy'
writing manifest file 'FiPy.egg-info/SOURCES.txt'
running build_ext
fipy version 3.1-dev74-gb542e0c
numpy version 1.10.1
pysparse is not installed
scipy version 0.15.1
matplotlib version 1.4.3
mpi4py is not installed
mpi4py is not installed
mayavi version 4.4.0
gmsh version 2.5.0
Traceback (most recent call last):
  File "setup.py", line 288, in <module>
    'Topic :: Software Development :: Libraries :: Python Modules'
  File "/Users/guyer/anaconda/envs/py2_7_debug/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/Users/guyer/anaconda/envs/py2_7_debug/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/Users/guyer/anaconda/envs/py2_7_debug/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/Users/guyer/anaconda/envs/py2_7_debug/lib/python2.7/site-packages/setuptools-18.4-py2.7.egg/setuptools/command/test.py", line 146, in run
[...]

PR #480, opened 2015-12-26
Issue476 setuptools 18 4 breaks test suite

guyer:
setuptools 18.0 changed how `test_args` and `test_suite` are used. This patch attempts to accommodate both the old and the new way.


guyer:
Addresses #476

Issue #482, opened 2016-01-26
FiPy development version string not compliant with PEP 440


guyer:
Recent versions of setuptools issue warnings like:
```
UserWarning: The version specified (u'3.1.1-dev126-g5c8209f') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
```

guyer:
Fixed by abc4f236d2587216337dcc8801ea888203e51302

PR #483, opened 2016-01-26
Format version string for pep 440

guyer:
Addresses #482


tkphd:
```python setup.py develop``` reports ```Finished processing dependencies for FiPy==3.1.dev130+gb0540c3```. New format looks PEP440-compliant to me.

tkphd:
New version strings render properly in PDF and HTML documentation, causing no build errors.

PR #489, opened 2016-02-09
Add travis build recipe

guyer:
Added a [Travis-CI](http://www.travis-ci.org) build recipe for automated testing.

Presently tests only on Mac OS X because that's where I had a complete `conda` stack. Once all prerequisites are available at anaconda.org, we should probably switch to a linux build as the [Travis-CI](http://www.travis-ci.org) resources are better there (doing both is not viable due to throttling if you ask for too many builds). [AppVeyor](https://www.appveyor.com) looks like  good option for CI on Windows.


guyer:
Failure of checks is due to problems in FiPy, not problems in this pull request (IMO)

tkphd:
Test results differ in serial and parallel environments, but this build recipe only tests the parallel side. Can we add serial builds without exceeding the throttling threshold?

guyer:
Build recipe tests serial and parallel. `MPIRUN` environment variable is empty for most build configurations. Total build matrix is 

- Python 2.7
  - Serial
    - PySparse
    - SciPy
    - Trilinos
  - Parallel (2 cores)
    - Trilinos
- Python 3.4
  - Serial
    - SciPy

tkphd:
Oh, so it is. OK, the Travis-CI configuration looks good to me, and the fact that the builds fail mean it's working as intended.

wd15:
@guyer, I'm assuming that you don't want this merged yet as you're still working on this.

PR #490, opened 2016-03-28
Preliminaries for Python 3 support

woodscn:
This PR lays groundwork for introducing Python3 support to FiPy. It heavily relies on the `future` package to generate code that is simultaneously compatible with both Python 2 and Python 3. The conversion, at this stage, is not complete. I have verified the basic tests, though I did not check those which required optional dependencies. 

At this point, 4 of the fixes in the future package have not yet been applied:
`libfuturize.fixes.fix_future_builtins`
`libfuturize.fixes.fix_future_standard_library`
`libfuturize.fixes.fix_future_standard_library_urllib`
`libfuturize.fixes.fix_division_safe`

A complete list of available fixes in the futurize script can be obtained by running:
`futurize -l`

Specific fixes can be applied using e.g.
`futurize -f libfuturize.fixes.fix_future_builtins`

This is as far as I can take the project, though. Hopefully it provides a good base for future work.

Edit: I should note that this PR (and any other which would provide Python 2/3 compatibility) introduces a runtime dependency on the `future` package. 

@pya @guyer


woodscn:
I'm not really sure why Travis is failing; everything passes on my machine when run with
`python setup.py test`

fipy/meshes/gmshMesh.py
```diff
@@ -835,7 +835,7 @@ def read(self):
 
             self.physicalFaceMap = nx.zeros(facesToV.shape[-1:], 'l')
             self.geometricalFaceMap = nx.zeros(facesToV.shape[-1:], 'l')
-            for face in facesDict.keys():
+            for face in list(facesDict.keys()):
```
guyer: These changes to `for foo in list(bar.keys()):` seem strange and non-idiomatic. The [cheat-sheet for futurize](http://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items) says that `for foo in bar:` is the proper way to write this in Python 2 and 3.

guyer:
The travis failure seems to be because travis isn't set up properly, yet. Let's not worry about that for the moment.

fipy/tools/dimensions/physicalField.py
```diff
@@ -1622,8 +1622,8 @@ def __rtruediv__(self, other):
         if isinstance(other,PhysicalUnit):
             return PhysicalUnit(other.names-self.names,
                                 other.factor/self.factor,
-                                map(lambda a,b: a-b,
-                                    other.powers, self.powers))
+                                list(map(lambda a,b: a-b,
```
guyer: This seems weird. Most of the other changes in this commit involve changing a map to a generator, which makes sense. I don't think explicitly casting a `map` to a `list` is the same thing. I wonder why these aren't just generators, too.

wd15:
This is too much for me right now.

tkphd:
@guyer will address this issue by documenting the technical benefits of Python 2 over Python 3, at which time this PR will be closed.

guyer:
Fixed by #645

PR #495, opened 2016-06-07
Memorize the C string representation.

cjgillot:
For some cases, the computation of the C
string for inline using _getRepresentation
takes a substantial amount of time.

Here, we memorize the computed C code and argument dictionary.
The argDict used to store the values of the required variables.
Now it stores the variables themselves,
and the values are retrieved by _execInline.

I have been using this patch offline for a while,
and it does not seem to harm anything.
What is your opinion about this ?

### DIFF
--- a/fipy/variables/variable.py
+++ b/fipy/variables/variable.py
@@ -386,6 +386,26 @@ class Variable(object):
             else:
                 return '[i + j * ni + k * ni * nj]'
 
+    def _getCvalue(self):
+         """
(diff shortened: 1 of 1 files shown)


wd15:
@cjgillot, thanks for this, we currently don't have the time to look into this pull request, but in the next round of FiPy development we will likely merge try and merge it.

wd15:
This might be too difficult for me to get into right now. I think I'll take myself off this.

PR #497, opened 2016-06-22
Add documented arguments to LinearBicgstabSolver

### LINKED ISSUE #496: scipy.LinearBicgstabSolver doesn't take arguments
As reported in [on the mailing list](http://comments.gmane.org/gmane.comp.python.fipy/4167), LinearBicgstabSolver doesn't take the arguments that it should (and that its docstring says it does).

Bug introduced in 4ea64a65

guyer:
Fixes #496


wd15:
@guyer, this seems to solve a small problem without failing while passing the tolerance argument to the solver so I'm going to merge it.

Issue #518, opened 2017-09-26
periodic and parallel are incompatible


guyer:
As [reported on the mailing list](https://www.mail-archive.com/fipy@nist.gov/msg04081.html), each "chunk" of a periodic grid is periodic, rather than being periodic between the bottom-most and top-most faces.

guyer:
This turns out to be due to a couple of things:
 - [ ] `mesh.facesTop`, etc. are calculated locally, not globally
   - [ ] PyTrilinos' `MaxAll` appears to be broken (it returns a list of values, instead of the global maximum)
 - [ ] the assertions to check the compatibility of connected faces can't work in parallel
   - [ ] faces from node 0 don't know about faces from node N
   - [ ] ghost cells aren't generated for periodic faces

To fix:
 - [ ] gather global faces
 - [ ] figure out compatible faces
   - [ ] check face centers colinear with other face normals
   - [ ] check faces parallel
   - note: this order, for common case of matching top and bottom of grid, will be O(N(N+1)), whereas reverse order will be O(2 N**2) (all faces are parallel, so all must be checked colinear, whereas typically only one other face will be colinear, so only it needs be checked parallel)
  - [ ] confirm areas match (combine with parallel check using current assertion on `_areaProjections`?)
 - [ ] create ghost cells
 - [ ] fix topology
 - [ ] fix geometry

PR #520, opened 2017-09-26
Make max/min global

### LINKED ISSUE #519: mesh.facesTop, etc. are determined on local mesh, not global
This script, when run in parallel, colors the top row of cells on every node.

```python
import fipy as fp

N = 10
mesh = fp.Grid2D(nx=N, ny=N)

top_faces = fp.FaceVariable(mesh=mesh, value=mesh.facesTop * 1.)
top_cells = top_faces.divergence

viewer = fp.Viewer(vars=top_cells)
viewer.plot()
```

This is because `.facesTop`, et al. calculate `max()` or `min()` over local faces, not global.

By and large, this doesn't matter, because ghost cells overwrite any constraints obtained this way, but as seen in #518, it can cause problems sometimes and is certainly not desired.

guyer:
FIxes #519

### DIFF
--- a/fipy/meshes/abstractMesh.py
+++ b/fipy/meshes/abstractMesh.py
@@ -724,7 +724,7 @@ class AbstractMesh(object):
 
         """
         x = self.faceCenters[0]
-        return x == _madmin(x)
+        return x == self._madmin(x)
 
     @property
     def facesRight(self):
@@ -744,7 +744,7 @@ class AbstractMesh(object):
 
         """
         x = self.faceCenters[0]
-        return x == _madmax(x)
+        return x == self._madmax(x)
 
     @property
     def facesBottom(self):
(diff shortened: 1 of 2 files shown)


fipy/meshes/abstractMesh.py
```diff
@@ -744,7 +744,7 @@ def facesRight(self):
 
         """
         x = self.faceCenters[0]
-        return x == _madmax(x)
+        return x == self._madmax(x)
```
tkphd: I would prefer a more descriptive function name than ```madmax```. Perhaps ```safemin``` and ```safemax```?

guyer: I've renamed them `_minmin` and `_maxmax` to denote that they return the minimum of all minima and the maximum of all maxima across processes.

guyer:
I'm withdrawing this pull request. Because of #518, this change will start raising exceptions for parallel periodic meshes. The exceptions are legitimate, but since the bug this change affects (#519) is masked for everything but parallel periodic, this fix should be integrated for that broader fix.

Issue #426, opened 2014-10-24
Robin boundary conditions have issues


guyer:
As discussed on the [mailing list](http://thread.gmane.org/gmane.comp.python.fipy/3638), there seem to be some problems with Robin boundary conditions.
- contraints don't seem to automatically update, e.g., 
  
  ``` python
  phi.faceGrad.constrain(-hh * (phi.faceValue - T0)/k, mesh.facesRight)
  while res > restol:
     res = eq.sweep(var=phi)
  ```
  
  behaves differently from 
  
  ``` python
  while res > restol:
     phi.faceGrad.constrain(-hh * (phi.faceValue - T0)/k, mesh.facesRight)
     res = eq.sweep(var=phi)
  ```
- `underRelaxation` may be necessary and should be illustrated, if so
-  Robin example [robin.py](../blob/8efb4378648c339a710ce021fb5e0c25a1899b4e/examples/convection/robin.py) seems to have issues:
  - `C.faceGrad.constrain([-P + P * C.faceValue], mesh.facesLeft)` will cast away `Variable`ness because a list containing a FiPy `Variable` is not a `Variable`. I think this should be `C.faceGrad.constrain((-P + P * C.faceValue) * mesh.faceNormals, mesh.facesLeft)` to properly assign the desired scalar value to the normal component of the gradient vector. Unfortunately, this change causes the numerical and analytical solutions to no longer agree (see below)
  - The analytical solution does not appear to be a solution to the stated equation and boundary conditions.

nlooije:
@guyer, @wd15: What is the status of this bug?

I am running into this problem in a 1D diffusion problem with a reacting boundary condition. Updating the boundary condition at every sweep drives the system to the correct solution but it is taking much longer than i feel is necessary.

wd15:
@nlooije, thanks for the heads up that this is an issue for you. I'm not sure when we can address it right now.

nlooije:
@wd15 - I was able to solve the problem using the method you [proposed](https://github.com/usnistgov/fipy/issues/426#issuecomment-61322554). I have scaled this up to a 2D convection-diffusion problem with reacting boundary conditions and it is working nicely. No longer updating the boundary condition at every sweep has sped up the simulation considerably. Thx!

[...]

PR #532, opened 2017-11-20
Document boundary conditions

guyer:
Add documentation for Robin and internal boundary conditions, [which](https://www.mail-archive.com/fipy@nist.gov/msg03671.html) [keeps](https://stackoverflow.com/questions/42769749/general-boundary-conditions) [coming](https://www.mail-archive.com/fipy@nist.gov/msg04091.html) [up](https://stackoverflow.com/questions/47169385/how-to-implement-internal-neumann-and-robin-conditions7) (see usnistgov/fipy#426, too)


documentation/USAGE.rst
```diff
@@ -535,68 +535,223 @@ Further demonstrations of spatially varying boundary condition can be found
...
-Applying internal boundary conditions
-=====================================
+Applying Robin boundary conditions
+==================================
+
+The Robin condition
+
+.. math::
+
+   a\phi + b\hat{n}\cdot\nabla\phi = g\qquad\text{on $f=f_0$}
```
wd15: Probably use \vec{a} there.

documentation/USAGE.rst
```diff
@@ -535,68 +535,223 @@ Further demonstrations of spatially varying boundary condition can be found
...
+
+Internal Robin condition
+------------------------
+
+Nothing different needs to be done when 
+`applying Robin boundary conditions`_ at internal faces.
+
+.. note::
+
+   While we believe these derivations are "correct", they often do not
```
wd15: Which derivations and what doesn't work? I'm not sure what you're referring to. Do you mean that "While the derivations above work well for external boundary conditions they fail to work for internal boundary conditions"?

documentation/USAGE.rst
```diff
@@ -535,68 +535,223 @@ Further demonstrations of spatially varying boundary condition can be found
...
+   seem to produce the intuitive result. At this point, we think this has 
+   to do with the pathology of "internal" boundary conditions, but remain 
+   open to other explanations. :term:`FiPy` was designed with diffuse 
+   interface treatments (phase field and level set) in mind and, as such, 
+   internal "boundaries" do not come up in our own work and have not 
+   received much attention.
+
+.. note::
+
+  One must be careful to distinguish between constraining internal cell
```
wd15: Maybe start off with what the note is actually about and/or make it a warning.

Warning: the constraints mechanism are not designed to constrain internal values for variables that are being solved by equations. In particular, one must be careful

documentation/USAGE.rst
```diff
@@ -535,68 +535,223 @@ Further demonstrations of spatially varying boundary condition can be found
...
->>> eqn.solve(var)
->>> print var
-[ 0.25  0.75]
+  >>> eqn.solve(var)
+  >>> print var
+  [ 0.25  0.75]
 
-However, if a constraint is used without the source term constraint an
-unexpected value is obtained
+  However, if a constraint is used without the source term constraint an
```
wd15: This is a little confusing, maybe layout two code blocks with all the code for both versions, but with no text interspersed

wd15:
Tests work and the documents build, it looks nice. Just a few changes.

documentation/USAGE.rst
guyer: Per [wikipedia](https://en.wikipedia.org/wiki/Robin_boundary_condition), it's a scalar equation. For completeness, it should probably be $\hat{n}\cdot\vec{a}$

documentation/USAGE.rst
guyer: The derivation for how to modify a DiffusionTerm and how to modify a ConvectionTerm. Oh, I see. I didn't mean that to be under "Internal Robin condition". It's a generic issue: I couldn't make that substitute-and-extrapolate approach work for `examples/convection/robin.py`, but I think the generic idea is on the right track for a Robin condition that isn't "just the flux".

documentation/USAGE.rst
guyer: agreed

documentation/USAGE.rst
guyer: Not sure I agree. I'm a strong believer in literate programming, but I'll look at making sure the presentation is as clear as possible.

documentation/USAGE.rst
```diff
@@ -535,68 +535,223 @@ Further demonstrations of spatially varying boundary condition can be found
...
-``value``. It does not have any effect on the implicit value of ``var`` at the
-``mask`` location during the linear solve so it is not a substitute
-for the source term machinations described above. Future releases of
-:term:`FiPy` may implicitly deal with this discrepancy, but the current
-release does not. A simple example can be used to demonstrate this::
+  simply fixes the returned value of ``var`` at ``mask`` to be
+  ``value``. It does not have any effect on the implicit value of ``var`` at the
+  ``mask`` location during the linear solve so it is not a substitute
+  for the source term machinations described above. Future releases of
+  :term:`FiPy` may implicitly deal with this discrepancy, but the current
```
tkphd: If there is not an issue for fixing this, please open one, or remove the time-dependence of the explanation.

documentation/USAGE.rst
guyer: Converted $a$ to $\hat{n}\cdot\vec{a}$ throughout

documentation/USAGE.rst
guyer: Clarified that the earlier Robin derivation is believed "correct"

documentation/USAGE.rst
```diff
@@ -535,68 +535,223 @@ Further demonstrations of spatially varying boundary condition can be found
...
+
+The Robin condition can be used to substitute for the expression 
+:math:`\left(\hat{n}\cdot\nabla\phi\right)_{f_0}`
+but we note that :term:`FiPy` calculates variable values at cell centers 
+and gradients at intervening faces. We obtain a first-order approximation 
+for :math:`\left(\hat{n}\cdot\nabla\phi\right)_{f_0}` in terms of 
+neighboring cell values by substituting 
+:math:`\phi_{f_0}
+\approx \phi_P - \left(\vec{d}_{fP}\cdot\nabla\phi\right)_{f_0}
+\approx \phi_P - \left(\hat{n}\cdot\nabla\phi\right)_{f_0}\left(\vec{d}_{fP}\cdot\hat{n}\right)_{f_0}`
```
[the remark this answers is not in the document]
guyer: all equations should be inline

documentation/USAGE.rst
guyer: Filed #581

documentation/USAGE.rst
guyer: @wd15 I've added some clarify text in f5273cc. Please check.

PR #556, opened 2018-02-08
revise dedication to the public domain

tkphd:
Updates language in LICENSE.rst to reflect the [current guidelines](https://www.nist.gov/director/licensing). Sets questions of Fed/FGR/Postdoc aside, which clarifies the licensing discussion for NIST-internal contributors.

---
This change is [Reviewable](https://reviewable.io/reviews/usnistgov/fipy/556)

### DIFF
--- a/LICENSE.rst
+++ b/LICENSE.rst
@@ -2,22 +2,36 @@
 Terms of Use
 ------------
 
-This software was developed at the `National Institute of Standards
-and Technology`_ by employees of the Federal Government in the course
-of their official duties.  Pursuant to `title 17 section 105`_ of the
-United States Code this software is not subject to copyright
-protection and is in the public domain.  FiPy is an experimental
+This software was developed by employees of the `National Institute
+of Standards and Technology`_, an agency of the Federal Government.
+Pursuant to `title 17 section 105`_ of the United States Code,
+works of NIST employees are not not subject to copyright
+protection, and this software is in the public domain.  FiPy is an experimental
 system.  NIST_ assumes no responsibility whatsoever for its use by
 other parties, and makes no guarantees, expressed or implied, about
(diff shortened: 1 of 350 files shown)


tkphd:
This PR appears incomplete. It changes the text in LICENSE.rst, but the dedication is mirrored in each source file. Withdrawn to revise all license blocks.

wd15:
@tkphd, in my view, which might not be shared, it's best just to have text about the module rather than boilerplate license and authorship in each file as the  repository itself has a license file and the source code tracker takes care of who wrote it and when etc. @guyer might not agree.

tkphd:
@wd15, I don't disagree, but the script to change all the boilerplate is done, and including a snippet vs. including the whole boilerplate is logically equivalent, so, I'm not strongly motivated to change it all again. I would be willing to if there's consensus that a snippet with pointer-to-LICENSE.rst is the right thing to do.

tkphd:
Changes made throughout. Reopening for review.

tkphd:
Based on [experience](https://github.com/YixuanTan/PMCGG), people will sometimes extract individual files for reuse; the distribution unit for FiPy should be considered the individual classes. Therefore, we should definitely not remove the public-domain dedication, or friendly request for attribution, since it provides potential future users with breadcrumbs to follow back to the original project.

I would be in favor of remodeling after the [CC0](https://creativecommons.org/2011/04/15/plaintext-versions-of-creative-commons-licenses-and-cc0/) blurb, *e.g.*

```
# FiPy is a work of National Institute of Standards and Technology employees, and is
# not subject to copyright protection. A copy of its dedication to the public domain
# accompanies the source code as LICENSE.rst, originally available for download from
# https://github.com/usnistgov/fipy.
```

PR #570, opened 2018-05-04
Fix output when PyTrilinos or PyTrilinos version is unavailable

shwina:


wd15:
@shwina, I approve these changes.

@guyer, I don't seem to be able to add myself or anyone else as a reviewer on this.

tkphd:
While the suggested fix does work, it seems to me that a better approach would fold PyTrilinos into the list of packages to be imported on [line 154](https://github.com/usnistgov/fipy/blob/43bbbd65ea1867fe73b0a58e00834aec5a616dc4/fipy/tests/testClass.py#L154), and deleting lines 170 through 178.

@guyer or @wd15, can you comment on whether PyTrilinos needs to have a separate import check?

tkphd:
```python
>>> import PyTrilinos
>>> print PyTrilinos.__version__
12.3;(;Dev;)
```
@wd15 seems to me that it could simply be added to the list, though I'm not sure how many versions back include `__version__`. I don't disagree with @guyer, though: the elegance of this solution is not important.

Issue #513, opened 2017-07-24
Stokes example gives wrong equation


guyer:
examples/flow/stokesCavity.py lists the "Navier-Stokes equation in the viscous limit,
$\nabla \mu \cdot \nabla \vec{u} = \nabla p$", but this should be $\nabla \cdot \left( \mu \nabla \vec{u}\right) = \nabla p$

PR #579, opened 2018-06-27
Update stokesCavity.py

Rowin:
If I understand well, pressure on line 170 shouldn't be a vector. This is a little one but can help understanding.

I'm not used to make pull request, so please tell me if I'm make it the wrong way.

### DIFF
--- a/examples/flow/stokesCavity.py
+++ b/examples/flow/stokesCavity.py
@@ -167,7 +167,7 @@ obtain,
 
 .. math::
 
-   \nabla \mu \cdot \nabla \vec{u}' = \vec{p}'
+   \nabla \mu \cdot \nabla \vec{u}' = p'
 
 and
 


tkphd:
Hey @Rowin, thanks for the pull request! This is how you're supposed to do it.

Since `u` is a vector, `\nabla \mu \cdot \nabla \vec{u}'` is also a vector, therefore `\vec{p}'` is the correct right-hand side. The equation stands as-is.

wd15:
@Rowin, we should perhaps use tensor notation to make this clearer. In tensor notation, this becomes

`\partial_j \left( \mu \partial_j u_i \right) = p_i`

with an implied summation over `j`. This better indicates how the indices match up IMO.

tkphd:
Oh, I see the error. It's related to #513, but the [transformed equation](https://github.com/usnistgov/fipy/blob/71c48d9037e5c3c1f8f3222b7f7a86a8bb8e5533/examples/flow/stokesCavity.py#L170) should be
```
\nabla \cdot \left( \mu \cdot \nabla \vec{u'} \right) = \nabla{p'}
```
which brings the substitutions leading to [L184](https://github.com/usnistgov/fipy/blob/71c48d9037e5c3c1f8f3222b7f7a86a8bb8e5533/examples/flow/stokesCavity.py#L184) into harmony.

Issuing a pull request to correct.

Issue #590, opened 2018-08-29
No module named pyAMGSolver


guyer:
[edit: I was mixing up `pyAMG` (ancient) and `pyamgx` (new)]

`fipy.solvers.pyAMG.smoothedAggregationSolver` [makes an undefined import](https://github.com/usnistgov/fipy/blob/develop/fipy/solvers/pyAMG/smoothedAggregationSolver.py#L53).

I can't tell whether `SmoothedAggregationSolver` needs a different base class, if there's something missing from the module, or if this bit of code isn't functional.

wd15:
@guyer, I can't really remember. Maybe there were redundancies between the pyamg and scipy solver classes so I was trying to merge them? I probably just overlooked the SmoothedAggregationSolver. I'll try and fix it then.

wd15:
@guyer, actually, I don't think the SmoothedAggregationSolver is needed since there is SmoothedAggregationPreconditioner which is actually what is used. I'm just going to delete it.

PR #593, opened 2018-08-30
Remove SmoothedAggregationSolver

wd15:
Address #590

The SmoothedAggregationSolver was accidentally left behind when PyAMG
was set up to use the Scipy solvers already in Scipy, see
0e630cac55e34d319. It's not used anywhere and was replaced by the
SmoothedAggregationPreconditioner.


wd15:
@guyer, do you want to be in charge of merges as you mostly know what's going on with FiPy?

guyer:
@wd15: I suppose so
@tkphd: awaiting your review

PR #596, opened 2018-09-02
Explain where to get examples

### LINKED ISSUE #533: pip or conda installation don't make clear where to get examples
Hello,

I'm new to FiPY and I tried installing it using pip install. However after the installation process is completed, I can't find the "Examples" folder in the directory. Is it normal or have I gotten it wrong? How do I find and run the examples codes otherwise?

Thanks for your help!

Dio

guyer:
Fixes #533

### DIFF
--- a/INSTALLATION.rst
+++ b/INSTALLATION.rst
@@ -93,6 +93,12 @@ Recommended Method
        sparse matrix :ref:`SOLVERS`, but the system still should be
        functional.
 
+.. attention::
+
+   When installed via conda_ or :term:`pip`, :term:`FiPy` will not include
+   its :ref:`examples <part:examples>`.  These can be obtained by 
+   `cloning the repository`_ or downloading a `compressed archive`_.
+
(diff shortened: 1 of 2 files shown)


wd15:
@dtwright03, what command did you use to clone the repository?

guyer:
@dtwright03 I don't think that's our `examples\` directory. Look in whatever directory you were in when you ran `git clone https://github.com/usnistgov/fipy.git`. There should be a `fipy\` directory and a `fipy\examples\` subdirectory.

dtwright03:
I activated the environment I wanted and ran git clone https://github.com/usnistgov/fipy.git.  Went smoothly.  I fired up spyder and went to the directory ...anaconda2\envs\fipyenv\Lib\site-packages\fipy and still no examples subdirectory.

Something probably only interesting to me is that looking at the code at usnitgov/fipy, it seems like the examples live in the same directory as, rather than a subdirectory of, fipy?

PR #634, opened 2019-04-10
Issue623 start using versioneer

### LINKED ISSUE #623: Start using versioneer
We should use [versioneer](https://github.com/warner/python-versioneer) for versioning

guyer:
Fixes #623


guyer:
My guess is the idea is to make sure it *always* works, but I agree that it's kind of weird.  Lot's  of projects using it though,  and it resolves some fragile stuff we were doing

wd15:
If Versioneer is no longer a dependency than it is removing a build time dependency, which is a good thing. I think it's because the version is something that is definitely needed at build time.

Issue #565, opened 2018-04-05
pip install does not work on empty env


achennu:
The [recommended method](https://www.ctcms.nist.gov/fipy/INSTALLATION.html#recommended-method) of installing works ok, for fipy as the main installation target.

Essentially it creates a python environment that already contains numpy, so `pip install fipy` works. However, numpy seems to be a setup time dependency for fipy, but it is not explicitly stated and so installing to an empty environment fails.

To demonstrate:

```console
conda create -n tmp python=2.7
source activate tmp
```
Now if I try:
```console
pip install --no-cache-dir fipy
# same for pip install --no-cache-dir fipy numpy
Collecting fipy
  Downloading FiPy-3.1.3.tar.gz (5.6MB)
    100% |████████████████████████████████| 5.6MB 13.8MB/s 
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-BzcLMs/fipy/setup.py", line 39, in <module>
        from fipy.tools.performance.efficiency_test import Efficiency_test
      File "fipy/__init__.py", line 45, in <module>
        from fipy.boundaryConditions import *
      File "fipy/boundaryConditions/__init__.py", line 2, in <module>
        from fipy.boundaryConditions.fixedFlux import *
      File "fipy/boundaryConditions/fixedFlux.py", line 38, in <module>
        from fipy.tools import numerix
      File "fipy/tools/__init__.py", line 85, in <module>
        serial, parallel = serialComm, parallelComm = _getComms()
      File "fipy/tools/__init__.py", line 77, in _getComms
        from fipy.tools.comms.dummyComm import DummyComm
      File "fipy/tools/comms/dummyComm.py", line 37, in <module>
        from fipy.tools import numerix
      File "fipy/tools/numerix.py", line 76, in <module>
        import numpy as NUMERIX
    ImportError: No module named numpy

```

Seems to be the case for the git repo install too still:

```console
pip install --no-cache-dir git+https://github.com/usnistgov/fipy#develop
Collecting git+https://github.com/usnistgov/fipy#develop
  Cloning https://github.com/usnistgov/fipy to /tmp/pip-Wh01Dc-build
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-Wh01Dc-build/setup.py", line 51, in <module>
[...]

PR #635, opened 2019-04-11
Issue625 setup py should not import fipy

### LINKED ISSUE #625: setup.py should not import fipy
[setup.py](https://github.com/usnistgov/fipy/blob/develop/setup.py#L4) imports several commands from the `fipy` package (`Efficiency_test`, `Copy_script`, `changelog`, `_TestClass`). As a result, simple commands like `python setup.py egg_info` don't work without at least some of FiPy's prerequisites, e.g., `numpy`, which is annoying, fragile, and potentially circular.

These commands should be moved either into `setup.py` or into an installation package that isn't `fipy`.

guyer:
Fixes #625


setup.py
```diff
@@ -3,161 +3,23 @@
 from distutils.core import Command
 import versioneer
```
wd15: This is importing the local versioneer.py version so versioneer isn't a dependency for installation.

guyer:
> There is no `install_requires` section in the `setup.py`. Should that be there? I think it's something that we've never had, right?

I intend to address `install_requires` with #565

guyer:
> Shouldn't the `getVersion` function in `setup.py` be dispensed with?

Gah! I made a complete hash when I rebased. I'll clean up and resubmit for review.

guyer:
> Recommend running pylint, flake8 and black...

Good thought. I'll let you know if I need help.

guyer:
Closing in favor of #637

Issue #624, opened 2019-03-13
Change README.rst to README.md


guyer:
GItHub has better rendering for Markdown than for ReST, especially for status badges. It's cosmetic, but it makes the README *much* more readable in GitHub, which is increasingly the way people see it.

Sphinx now has [Markdown support](https://www.sphinx-doc.org/en/master/usage/markdown.html) using [recommonmark](https://recommonmark.readthedocs.io).

guyer:
[recommonmark](https://recommonmark.readthedocs.io/) has issues:
- it doesn't respect either '  ' or '\' for [Markdown line breaks](https://www.markdownguide.org/basic-syntax#line-breaks)
- it doesn't correctly interpret the alt text in [Markdown images](https://www.markdownguide.org/basic-syntax#images-1)

guyer:
Pull #636 managed to address the underlying problems with rendering badges in different media that motivated this issue

PR #636, opened 2019-04-12
Make badges work in GitHub and pdf

guyer:
Addresses #624 

Markdown support has issues and the only real point was to get the badges to show up in the GitHub README


wd15:
Sorry, Depsy is active, but no longer maintained. It's not important.

PR #637, opened 2019-04-13
Issue625 setup py should not import fipy bis

### LINKED ISSUE #625: setup.py should not import fipy

guyer:
Fixes #625
Supercedes #635

pylint gives 8.89/10.  Remaining complaints are stupid.


guyer:
@wd15 @tkphd Unlike #638, this one is critical path. Can I get a review, please?

wd15:
BTW, can make flake8 comply with black by adding
```
[flake8]
ignore = E731, W503
max-line-length = 88
```
to setup.cfg

wd15:
Why not add the linters to .travis.yml at this point? Just for setup.py for now.

guyer:
> Why not add the linters to .travis.yml at this point? Just for setup.py for now.

Added linting to circle because it's faster than travis and easier to compartmentalize

PR #642, opened 2019-04-21
Fix ADMINISTRATA

guyer:
- Revise ADMINISTRATA  to  reflect  what actually works
- Introduce `release` command for `setup.py` that automates actions of building distributions.
- Move `MANIFEST.in` to `MANIFEST-UNIX.in` to avoid [versioneer](https://github.com/warner/python-versioneer) getting crabby about dirty working trees.


guyer:
Utlimately this should be automated (#643)

documentation/ADMINISTRATA.rst
```diff
@@ -241,34 +239,13 @@ Test the installed compressed distribution::
     $ cd ..
     $ \rm -rf tmp
```
tkphd: `\rm` --> `rm`

tkphd: I stand corrected: @guyer points out that `\` strips aliases, so `\rm` is a perfectly valid shell command.

PR #647, opened 2019-05-14
Fix sphinx documentation

### LINKED ISSUE #457: Special methods are not included in Sphinx documentation
Documentation and examples placed in the docstrings of [Python special methods](https://docs.python.org/2/reference/datamodel.html#special-method-names), e.g., `__init__`, `__add__`,  etc., is not included in our Sphinx documentation. We should either figure out how to include these, or probably better and more consistent with NumPy and SciPy, move this documentation to the class docstring.

See http://stackoverflow.com/a/32704639/2019542

### LINKED ISSUE #552: documentation integration
Continuous Integration service (*i.e.* [Travis](https://www.travis-ci.org)) should test documentation builds, since we do not intend to use a third-party documentation host (*e.g.* [readthedocs](https://www.readthedocs.io)).

Related to #457, #551.

### LINKED ISSUE #601: contents.rst and manual.rst are a recursive mess
https://www.ctcms.nist.gov/fipy/documentation/contents.html has duplicate entries because contents.rst includes manual.rst and they both pull in the some of the same files

guyer:
Address issues that prevent documentation from building and/or cause it to build poorly.

- Switch from [numpydoc](https://numpydoc.readthedocs.io) (external and fussy) to [napoleon](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) (included in sphinx)
- Update title page to current sphinx syntax
- Clean up document structure (fixes #601)
- Document `__special__` members (fixes #457)
- Add documentation to CircleCI (fixes #552)


examples/convection/exponential1D/mesh1D.py
```diff
@@ -17,7 +17,8 @@
 
 We define a 1D mesh
 
-.. index:: Grid1D
+.. index::
+   single: Grid1D
```
wd15: What is this supposed to do? Is something supposed to link from the General Index to this reference? I can't find that in the index. 

BTW, in my version of the document build the Global Module Index is a dead link from the the main index.html. That's the same on the version hosted on CTCMS.

documentation/USAGE.rst
```diff
@@ -959,4 +959,9 @@ command in the base directory::
 
    $ conda install --channel conda-forge matplotlib pandas imagemagick
 
+   The PDF file requires `SIunits.sty`_ available, e.g., from
+   `texlive-science`_.
+
 .. _download the latest manual:  http://www.ctcms.nist.gov/fipy/download/
+.. _SIunits.sty: https://ctan.org/pkg/siunits
```
tkphd: Per the linked [CTAN](https://ctan.org/pkg/siunits) page, SIunits is superseded by [SIunitx](https://ctan.org/pkg/siunitx). It's available for both TeXLive (Linux/MacOS) and MiKTeX (Windows).

guyer: I'm aware. It's presumably different in some way (if it's not different in some way, why did they "supercede" it?!?). Since `SIunits.sty` can still be obtained, I didn't feel that this particular pull request needed to delve into switching.

examples/convection/exponential1D/mesh1D.py
guyer: `.. index:: Grid1D` is a shorthand that napoleon doesn't understand and I didn't feel like fighting with. I think it may be a [sphinx augmentation](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#index-generating-markup).

I see what you mean about the Global Module Index. I'll look into that. Search is also broken, and has been for awhile. That seems to be an [issue with sphinx](https://github.com/sphinx-doc/sphinx/issues/5460) that may or may not be fixed in the version available for Python 2.

documentation/USAGE.rst
tkphd: Since SIunits is in maintenance mode, not entirely deprecated, I guess it's OK? Feels like pushing technical debt onto our future-selves, though.

PR #657, opened 2019-06-24
Automate spell check

guyer:
Uses [Pyspelling](https://facelessuser.github.io/pyspelling)


wd15:
This is really awesome. Can you set it up to handle different formatting types in the docstring?

PR #678, opened 2019-11-27
fix: use correct logo links

wd15:
This is a temporary fix for the FiPy website so it uses the same logo links as the nist.gov pages. The NIST / MML theme may need to be abandoned (or use a more recent version) and [this](https://github.com/usnistgov/nist-header-footer) used instead. I wasn't sure how it all works so will leave that to @guyer unless he delegates.


guyer:
Superseded by #680

PR #680, opened 2019-12-06
Fix nist header footer

guyer:
Replace hard-coded header/footer with reference to
[official template](https://github.com/usnistgov/nist-header-footer)

NIST Drupal got messed all around. They've provided an easy way to get look-and-feel headers and footers, but styling everything else has to be done manually with static copies. Ugh.


.circleci/config.yml
```diff
@@ -375,7 +375,7 @@ jobs:
           command: |
             source activate ~/project/test-environment-27
             conda install --channel conda-forge sphinx future
-            pip install sphinxcontrib-bibtex
+            pip install sphinxcontrib-bibtex <=0.4.2
```
wd15: I needed quotes around that in bash, pip install "sphinxcontrib-bibtex<=0.4.2"

guyer: Quotes were already added in a later commit, but thanks

PR #681, opened 2019-12-17
Fix link to Nick Croft's thesis

wd15:
Nick Croft's thesis link is updated and fixed (couldn't find a
DOI). Excessive white space also removed from refs.bib.

### DIFF
--- a/documentation/refs.bib
+++ b/documentation/refs.bib
@@ -25,7 +25,7 @@
               Turbulent Reacting Flows},
    school =  {University of Greenwich},
    year =    1998,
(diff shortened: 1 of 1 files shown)


tkphd:
The Greenwich Academy page indicates that the provided link is a URI, which *should* be equivalent to a DOI. The thesis also has a British Library EThOS record: [uk.bl.ethos.245782](https://ethos.bl.uk/OrderDetails.do?uin=uk.bl.ethos.245782), from which the Greenwich Academy's [URI](https://gala.gre.ac.uk/id/eprint/6371/) is linked.

I think the URI is the better link, since it goes directly to a downloadable artifact, though the EThOS link may be longer-lived.

Issue #692, opened 2020-01-19
Can't copy example scripts with the command line


awa5114:
Hey does anyone know why this occurs? I did a clean development install with python 3.6 exactly as outlined by https://www.ctcms.nist.gov/fipy/INSTALLATION.html#environment

When I try to copy one of the example scripts using the command line I get an error:

```
(fipy-dev) D:\github_projects\fipy>python setup.py copy_script --From examples/diffusion/mesh1D.py --To D:/CGF-1/notebooks/fipy_examples/mesh1D.py
running copy_script
The file 'D:/CGF-1/notebooks/fipy_examples/mesh1D.py' already exists. Overwrite? [n] y
Traceback (most recent call last):
  File "setup.py", line 87, in <module>
    "Topic :: Software Development :: Libraries :: Python Modules",
  File "C:\Users\aboufira\AppData\Local\Continuum\miniconda3\envs\fipy-dev\lib\site-packages\setuptools\__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "C:\Users\aboufira\AppData\Local\Continuum\miniconda3\envs\fipy-dev\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\Users\aboufira\AppData\Local\Continuum\miniconda3\envs\fipy-dev\lib\distutils\dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "C:\Users\aboufira\AppData\Local\Continuum\miniconda3\envs\fipy-dev\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "D:\github_projects\fipy\_setup\copy_script.py", line 57, in run
    f = file(self.To, "w")
NameError: name 'file' is not defined
```
Does this have anything to do with the fact that I am using python 3 instead of 2? Thanks for your help.

wd15:
Changing `file` to `open` works. I'll submit a PR for that change.

wd15:
@awa5114, but thanks for raising the issue!

PR #695, opened 2020-01-21
fix: use open instead of file in py3

wd15:
Address #692

Use the open function instead of the file function in Python 3. The
copy_script command still had the Python 2 syntax.


guyer:
Fixes #692

PR #698, opened 2020-01-27
Fix nist header footer

guyer:


guyer:
This was an accidental reapplication of #680. Only actual change was shifting CircleCI to build docs under Py3k.

PR #699, opened 2020-01-27
For benchmark 8

guyer:
Assorted patches to get [phase field benchmark 8](https://pages.nist.gov/pfhub/benchmarks/benchmark8.ipynb/) working.


fipy/variables/cellVariable.py
```diff
@@ -211,8 +211,8 @@ def cellVolumeAverage(self):
         """
 
         if not hasattr(self, 'volumeAverage'):
-            from fipy.variables.cellVolumeAverageVariable import _CellVolumeAverageVariable
-            self.volumeAverage = _CellVolumeAverageVariable(self)
+            volumes = CellVariable(mesh=self.mesh, value=self.mesh.cellVolumes)
```
wd15: Why are you creating another CellVariable here?

wd15:
Looks fine, but I don't think that an intermediate CellVariable is needed.

fipy/variables/cellVariable.py
guyer: It's been like that in `_CellVolumeAverageVariable` for about 11 years, but you're right

PR #700, opened 2020-01-27
Assorted fixes while supporting PETSc

guyer:


examples/diffusion/steadyState/mesh20x20/modifiedMeshInput.py
```diff
@@ -31,7 +32,7 @@
 ...     Line Loop(10) = {6, 7, 8, 9};
 ...
 ...     Plane Surface(11) = {10};
-...     ''') # doctest: +GMSH
+...     ''' % locals()) # doctest: +GMSH
```
wd15: I'm not a fan of the `% locals()` thing. It would be best to use f-strings, but they aren't available before Python 3.7. Just list the variable in keyword arguments to format.

guyer: We use it in 10 other places. Do you want them all gone?

wd15: Ok, ok if that's the way we've been doing it. I thought you'd just decided to do that.

PR #704, opened 2020-01-31
Ping

guyer:


guyer:
This was just check that @crhume and I had reconfigured AppVeyor correctly

PR #710, opened 2020-02-10
build(nix): fix broken plm_rsh_agent error

wd15:
Fix issue with path to SSH for MPI when firing up the Nix shell.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -29,6 +29,8 @@ in
       export PYTHONPATH=$PYTHONPATH:$USER_SITE
       export PATH=$PATH:$PYTHONUSERBASE/bin
 
+      export OMPI_MCA_plm_rsh_agent=/usr/bin/ssh
+
       ## To build the docs
       # pip install --user sphinx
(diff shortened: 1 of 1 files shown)


guyer:
What am I supposed to be looking at?

- `nix-shell --pure --command "python setup.py test"` runs
- `nix-shell --pure --command "mpirun -np 2 python setup.py test --petsc"` does not 
("`bash: mpirun: command not found`")

wd15:
> What am I supposed to be looking at?

Well, in this case `nix-shell --pure` didn't work at all for me after the petsc merge (I'm guessing). I was getting a great long `plm_rsh_agent` MPI error. No idea why this happens, but maybe mpi4py or something is actually imported even when not in parallel, which is fine, but maybe a change from before petsc.

Anyway, this change fixes it, but it's very fragile as it's pointing at a local SSH build. 

>     * `nix-shell --pure --command "python setup.py test"` runs

This too is good to check, obviously.

>     * `nix-shell --pure --command "mpirun -np 2 python setup.py test --petsc"` does not

Yes, I haven't got to petsc or any of that stuff yet.

PR #714, opened 2020-03-10
Try to fix the Dirichlet Boundary Condition

atismer:
See issue in mailing list
https://www.mail-archive.com/fipy@nist.gov/msg04371.html

### DIFF
--- a/fipy/terms/abstractConvectionTerm.py
+++ b/fipy/terms/abstractConvectionTerm.py
@@ -164,9 +164,10 @@ class _AbstractConvectionTerm(FaceTerm):
         mesh = var.mesh
 
         if (not hasattr(self, 'constraintL')) or (not hasattr(self, 'constraintB')):
-
-            constraintMask = var.faceGrad.constraintMask | var.arithmeticFaceValue.constraintMask
-
+            
+            constraintMaskFixedGradient = var.faceGrad.constraintMask
+            constraintMaskFixedValue = var.arithmeticFaceValue.constraintMask
+            
             weight = self._getWeight(var, transientGeomCoeff, diffusionGeomCoeff)
 
             if 'implicit' in weight:
@@ -176,9 +177,9 @@ class _AbstractConvectionTerm(FaceTerm):
 
             exteriorCoeff =  self.coeff * mesh.exteriorFaces
 
-            self.constraintL = (alpha * constraintMask * exteriorCoeff).divergence * mesh.cellVolumes
-            self.constraintB =  -((1 - alpha) * var.arithmeticFaceValue * constraintMask * exteriorCoeff).divergence * mesh.cellVolumes
-
+            self.constraintL = (constraintMaskFixedGradient * exteriorCoeff).divergence * mesh.cellVolumes
+            self.constraintB =  -(var.arithmeticFaceValue * constraintMaskFixedValue * exteriorCoeff).divergence * mesh.cellVolumes
+            
         ids = self._reshapeIDs(var, numerix.arange(mesh.numberOfCells))
         L.addAt(numerix.array(self.constraintL).ravel(), ids.ravel(), ids.swapaxes(0, 1).ravel())
         b += numerix.reshape(self.constraintB.value, ids.shape).sum(0).ravel()
--- a/fipy/terms/abstractConvectionTerm.py
+++ b/fipy/terms/abstractConvectionTerm.py
@@ -164,10 +164,9 @@ class _AbstractConvectionTerm(FaceTerm):
         mesh = var.mesh
 
         if (not hasattr(self, 'constraintL')) or (not hasattr(self, 'constraintB')):
-            
-            constraintMaskFixedGradient = var.faceGrad.constraintMask
-            constraintMaskFixedValue = var.arithmeticFaceValue.constraintMask
-            
+
+            constraintMask = var.faceGrad.constraintMask | var.arithmeticFaceValue.constraintMask
+
             weight = self._getWeight(var, transientGeomCoeff, diffusionGeomCoeff)
 
             if 'implicit' in weight:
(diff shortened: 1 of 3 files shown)


atismer:
I did a small validation of my code. In the book

> An Introduction to Computational Fluid Dynamics
> The Finite Volume Method
> second edition

of

> Versteeg and Malalasekera

is a small example (Example 5.1 pages 137-141).

I reproduced the Example in fipy with
decl
and I get the following result:
versteeg

The same code running with my small fix gives exactly the same results as in the book:
fix_versteeg

Additionally, I change the boundary condition from a fixed value at position x=1 to a zero gradient condition. Using fipy gives a strange and not a physical result. The value of phi increases. I think this makes no sense:
versteegMod

Running the modified example with my fixed code gives more physical results:
fix_versteegMod

I think that there is something wrong in the code. With my fix the example runs and is in line with the book. Additionally, the modified example gives better results with the fix. I hope that my fix and my small investigation is of any help for you.

It would be very nice if you give me a feedback.
Thanks.

guyer:
Please post a link to a gist of this notebook.

Please address the test failures that result from this pull request.

atismer:
> Please post a link to a gist of this notebook.

Link to the notebook on gist:
[https://gist.github.com/atismer/4448eb13496c53191a4f7052de2ab16d](https://gist.github.com/atismer/4448eb13496c53191a4f7052de2ab16d)

> Please address the test failures that result from this pull request.
I will have a look to the tests and I will come back to you.

guyer:
Thanks. We'll take a look.

atismer:
> Please address the test failures that result from this pull request.

For this commit the checks run successfully. Sorry for the first commit that does not.

Basically, I move the fix from _abstractConvectionTerm.py_ to _centralDiffConvectionTerm.py_. It might be not the cleanest solution, because of the call of **_AbstractConvectionTerm._buildMatrix** in **CentralDifferenceConvectionTerm._buildMatrix**. This makes it necessary to undo the changes in the matrix L and right hand side b (see line 44-47 in _centralDiffConvectionTerm.py_). I'm not sure if the call is necessary, but I was not successfull in calling the method **FaceTerm._buildMatrix** directly.

With this commit also the results of the example are in line with the book if Versteeg.
So I did my best. Hope this is of any help for you.

wd15:
Thanks for the pull request. Well done for getting into the depths of FiPy. I was involved with writing this code, but it's quite painful trying to remember how things worked.

I'm going to start with the first part of this problem with the left boundary condition constrained to 1 and the right constrained to 0. I'm running this code to test it

```
import fipy as fp
import numpy

nx = 100
dx = 1./nx
mesh = fp.Grid1D(nx=nx, dx=dx)

phi = fp.CellVariable(name="phi", mesh=mesh, value=0.)
phi.constrain(1.0, mesh.facesLeft)
phi.constrain(0.0, mesh.facesRight)
eq = fp.CentralDifferenceConvectionTerm(coeff=(0.1,)) - fp.DiffusionTerm(coeff=0.1) == 0
eq.cacheMatrix()
eq.cacheRHSvector()
eq.solve(var=phi,solver=fp.solvers.LinearLUSolver())

actual = phi.value
expected = (numpy.exp(mesh.x) - numpy.exp(1.)) / (1 - numpy.exp(1.))
error = actual - expected
print(numpy.sqrt(float((error**2).sum())))
```

Now with the master branch of FiPy I get an error of 7.17e-6, but with the new changes I get an error of 1.36e-4, which is considerably worse. As you point out your matrix agrees with the book, but it isn't more accurate when compared with the analytical result.

wd15:
For the second part, I still haven't fully understood the constraint issue, but the following does eventually get the correct result

```
import fipy as fp
import numpy

nx = 100
dx = 1./nx
mesh = fp.Grid1D(nx=nx, dx=dx)

phi = fp.CellVariable(name="phi", mesh=mesh, value=0.)
phi.constrain(1.0, mesh.facesLeft)
phi.faceGrad.constrain([0.0], mesh.facesRight)
eq = fp.CentralDifferenceConvectionTerm(coeff=(0.1,)) - fp.DiffusionTerm(coeff=0.1) == 0

for i in range(100):
    eq.sweep( var=phi,solver=fp.solvers.LinearLUSolver() )
    print(phi.value[-1])
```

Now, I think the issue is that the central difference convection uses a "natural" outflow condition when the gradient constraint is set. This was chosen for backwards compatibility with the way FiPy worked before constraints were implemented. Basically, as the initial value is 0.0 for this then the outflow condition thinks its value on the right face is 0 on the first iteration. If "phi" is set to 1. initially then it works on the first iteration. In essence with this formulation of the boundary conditions, iterations are required.

We'll need to think carefully if we want the default not to be a natural boundary condition in this case. I agree that this highly non-intuitive when you write down the equations and we don't get the correct result on the first iteration.

wd15:
I think that we mostly suggest the following to impose a specific outflow

```
import fipy as fp
import numpy

nx = 100
dx = 1./nx
mesh = fp.Grid1D(nx=nx, dx=dx)

phi = fp.CellVariable(name="phi", mesh=mesh, value=0.)
phi.constrain(1.0, mesh.facesLeft)
# phi.constrain(1.0, mesh.facesRight)
phi.faceGrad.constrain([0.0], mesh.facesRight)
convCoeff = fp.FaceVariable(mesh, 0.1, rank=1)
convCoeff[-1] = 0.0


eq = fp.CentralDifferenceConvectionTerm(coeff=convCoeff) - fp.DiffusionTerm(coeff=0.1) == 0

eq.sweep( var=phi,solver=fp.solvers.LinearLUSolver() )
print(phi.value[-1])
```

@guyer, I think this needs revisiting. I think we want to keep the accuracy that we have with the two fixed BCs (first result) and still get the correct result in one iteration for this problem. What do you think?

atismer:
> Now with the master branch of FiPy I get an error of 7.17e-6, but with the new changes I get an error of 1.36e-4, which is considerably worse. As you point out your matrix agrees with the book, but it isn't more accurate when compared with the analytical result.

This is correct. I agree with you.

> I agree that this highly non-intuitive when you write down the equations and we don't get the correct result on the first iteration.

This is correct and I think this solution would have been never come to my mind. But many thanks for the answer.

wd15:
Hopefully, there is a good resolution to this in #719

guyer:
Incorporated in #719

PR #725, opened 2020-05-05
Issue662 issues scaling colorbar with datamin and datamax args

### LINKED ISSUE #662: Issues Scaling Colorbar with Datamin and Datamax Args
When I run the code for "examples.diffusion.mesh20x20" (using Python 3.7) the viewer has a color bar scaled from -1 to 1, despite the line of code where the datamin and datamax are specified, i.e.,
viewer = Viewer(vars=phi, datamin=0., datamax=1.)

Any insights on how to scale the color bar effectively?
Diffusion_20x20

### LINKED ISSUE #663: datamin/datamax argument ignored by viewer
Hi,

Similarly to issue #662 the datamin/datamax argument is not accounted for by matplotlib on a new install of fipy 3.3 with python 3.6.9 using conda in Win7.

So whenever i use the viewer, the data range is set to [-1, 1] 

I am getting the following warning from matplotlib, not sure if it is related:

C:\..\Anaconda3\envs\fipy-3.3\lib\site-packages\fipy\viewers\matplotlibViewer\matplotlib
Viewer.py:229: MatplotlibDeprecationWarning:
The set_norm function was deprecated in Matplotlib 3.1 and will be removed in 3.3. Use ScalarMappable.set_norm instead.
  self._cb.set_norm(value)

Do i have to go back to using fipy with python 2 ?

guyer:
Fix a number of issues with colorbars that came with Matplotlib 3.

Turns out we were never using norms correctly. Hopefully better now.

Fixes #662. Fixes #663.

### DIFF
--- a/fipy/viewers/matplotlibViewer/matplotlib2DGridViewer.py
+++ b/fipy/viewers/matplotlibViewer/matplotlib2DGridViewer.py
@@ -47,14 +47,6 @@ class Matplotlib2DGridViewer(AbstractMatplotlib2DViewer):
                                             cmap=cmap, colorbar=colorbar, axes=axes, figaspect=figaspect,
                                             **kwlimits)
 
-        xmin, ymin = self.vars[0].mesh.extents['min']
-        xmax, ymax = self.vars[0].mesh.extents['max']
-
-        self.image = self.axes.imshow(self._data,
-                                      extent=(xmin, xmax, ymin, ymax),
-                                      vmin=0, vmax=1,
-                                      cmap=self.cmap)
-
         self.axes.set_xlim(xmin=self._getLimit('xmin'),
                            xmax=self._getLimit('xmax'))
 
@@ -90,19 +82,24 @@ class Matplotlib2DGridViewer(AbstractMatplotlib2DViewer):
         # this viewer can only display one variable
         return [vars[0]]
 
+    def _make_mappable(self):
+        xmin, ymin = self.vars[0].mesh.extents['min']
+        xmax, ymax = self.vars[0].mesh.extents['max']
+
+        self.image = self.axes.imshow(self._data,
+                                      extent=(xmin, xmax, ymin, ymax),
+                                      norm=self.norm,
+                                      cmap=self.cmap)
+        return self.image
+
     @property
     def _data(self):
         from fipy.tools.numerix import array, reshape
         return reshape(array(self.vars[0]), self.vars[0].mesh.shape[::-1])[::-1]
 
     def _plot(self):
-        self.norm.vmin = self._getLimit(('datamin', 'zmin'))
-        self.norm.vmax = self._getLimit(('datamax', 'zmax'))
-
-        self.image.set_data(self.norm(self._data))
-
-        if self.colorbar is not None:
-            self.colorbar.plot()
+        super(Matplotlib2DGridViewer, self)._plot()
+        self.image.set_data(self._data)
 
 def _test():
     from fipy.viewers.viewer import _test2D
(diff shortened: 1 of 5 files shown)


fipy/viewers/matplotlibViewer/matplotlib2DGridViewer.py
```diff
@@ -90,19 +82,24 @@ def _getSuitableVars(self, vars):
...
 
     def _plot(self):
-        self.norm.vmin = self._getLimit(('datamin', 'zmin'))
-        self.norm.vmax = self._getLimit(('datamax', 'zmax'))
-
-        self.image.set_data(self.norm(self._data))
-
-        if self.colorbar is not None:
-            self.colorbar.plot()
+        super(Matplotlib2DGridViewer, self)._plot()
```
tkphd, line 101: Haven't seen `super` before... [apparently an inheritance trick](https://stackoverflow.com/questions/222877/what-does-super-do-in-python) to help other classes use this one's `_plot` method?

fipy/viewers/matplotlibViewer/matplotlib2DViewer.py
```diff
@@ -4,18 +4,37 @@
...
-    def figaspect(self, figaspect):
+    def figaspect(self, figaspect, colorbar):
         if figaspect == 'auto':
             figaspect = self.vars[0].mesh.aspect2D
+            # We can't make the colorbar, yet (as we don't even have a figure)
+            # so hardcode these values, based on
+            # https://matplotlib.org/api/_as_gen/matplotlib.figure.Figure.html#matplotlib.figure.Figure.colorbar
+            #   fraction	0.15; fraction of original axes to use for colorbar
+            #   pad	        0.05 if vertical, 0.15 if horizontal; fraction of original axes between colorbar and new image axes
+            if colorbar == "vertical":
```
tkphd, line 22: Looks good, except using the variable name `colorbar` to represent an orientation, rather than something more intuitively colorbar-related, seems awkward.

fipy/viewers/matplotlibViewer/matplotlib2DGridViewer.py
wd15: I don't think that's what super does. I may misunderstand what you intended with your statement though. Super allows the sub-class (where it's used) to not require reference to the parent class by name. If the inheritance tree changes or the method changes position in the inheritance (it might move up the tree for example) tree then there's no need to change the code that calls the inherited method. That's the main benefit.

wd15: That's actually subtly wrong. I should amend that to say if the parent class name changes then there's no need to change all the calls to the parent class method. That is all it does.

wd15:
Fixes things for me. This looks good now,

```
from fipy import Grid2D, CellVariable, Viewer
# from fipy.viewers.matplotlibViewer.matplotlibViewer import _ColorBar

m = Grid2D(nx=3, ny=3)

v = CellVariable(mesh=m)
v[:] = m.x * m.y

vi = Viewer(v)
#vi.colorbar = _ColorBar(viewer=vi, vmin=0.0, vmax=9.0)

vi.plot()

input('stopped')
```

It would be really nice if we had notebook examples and we were testing them with nbval. I need to work on that at some point.

fipy/viewers/matplotlibViewer/matplotlib2DViewer.py
guyer: What should it represent? What should it be called?

I don't think an overdue bug fix is the place for a change to a 10 year old API.

PR #732, opened 2020-06-02
Spherical1D (Uniform and NonUniform) meshes

klkuhlm:
Based on the suggestion in this stack overflow question:

https://stackoverflow.com/questions/52242891/1d-spherical-grid-in-fipy

I have created two simple 1D spherical meshes (uniform and non-uniform), by adding an extra "r" term in the area calculations of the 1D cylindrical mesh.

Aside from trivial naming differences, the diff between the uniform cylindrical and spherical versions mainly is:

```
37c37
<         return self.faceCenters[0].value
---
>         return self.faceCenters[0].value * self.faceCenters[0].value
57c57
<         return self.dx * self.cellCenters[0].value
---
>         return self.dx * self.cellCenters[0].value * self.cellCenters[0].value

```

### DIFF
--- a/fipy/meshes/factoryMeshes.py
+++ b/fipy/meshes/factoryMeshes.py
@@ -5,7 +5,7 @@ __docformat__ = 'restructuredtext'
 from fipy.tools import parallelComm
 from fipy.tools import numerix
 
-__all__ = ["Grid3D", "Grid2D", "Grid1D", "CylindricalGrid2D", "CylindricalGrid1D"]
+__all__ = ["Grid3D", "Grid2D", "Grid1D", "CylindricalGrid2D", "CylindricalGrid1D", "SphericalGrid1D"]
 from future.utils import text_to_native_str
 __all__ = [text_to_native_str(n) for n in __all__]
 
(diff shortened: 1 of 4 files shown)


guyer:
@klkuhlm  Thank you very much for the contribution!

Presently, the test suite is showing singular matrix failures with the `scipy` solvers. If I had to guess, this is because there's a face with zero area at r = 0, but I don't know why this wouldn't fail with all solvers.

Can you try to diagnose what's going on here?

klkuhlm:
I have found I can trigger what I think is the same issue on my system (where all tests pass), by specifying the solver to be the LinearLU (which is maybe the default for scipy?).  

The snippet below triggers for the spherical1D mesh, but not the cylindrical1D mesh

I will try to investigate this further, but I wonder if maybe this means more to you than it does to me

```
# coding: utf-8
import fipy
Sm = fipy.meshes.SphericalGrid1D(nx=3.,dx=1.)
Cm = fipy.meshes.CylindricalGrid1D(nx=3.,dx=1.)
Sv = fipy.CellVariable(mesh=Sm)
Cv = fipy.CellVariable(mesh=Cm)
fipy.DiffusionTerm().solve(Sv)
fipy.DiffusionTerm().solve(Cv)
fipy.DiffusionTerm().solve(Sv,solver=fipy.solvers.LinearLUSolver())
fipy.DiffusionTerm().solve(Cv,solver=fipy.solvers.LinearLUSolver())
```

/Users/klkuhlm/anaconda3/envs/fipy-dev/lib/python2.7/site-packages/FiPy-3.4.1+6.g1f7c4be3.dirty-py2.7.egg/fipy/solvers/pysparse/linearLUSolver.py:72: RuntimeWarning: invalid value encountered in double_scalars
  if (numerix.sqrt(numerix.sum(errorVector**2)) / error0)  <= self.tolerance:

klkuhlm:
I guess what I posted is just a warning, not an error.  Still looking at it

klkuhlm:
now that I look closer that previous warning has nothing to do with anything.

guyer:
Yeah, the warning is just a reflection that the initial condition is an exact solution.

The issue is that the problem admits an infinite number of solutions. I don't know why scipy LU is less tolerant than LU from other solvers, nor why we don't see the same issue with the cylindrical meshes, but applying
```python
Sv.constrain(0., where=mesh.facesRight)
```
seems to resolve it.

klkuhlm:
thanks, that suggestion fixed it. The remaining failed travis checks appear unrelated.

klkuhlm:
I don't understand why the nearly identical cylindrical mesh doesn't trigger this same error. I triple-checked the differences between the two meshes and didn't see anything. 

I compared 1D cartesian, cylindrical, and spherical meshes for a diffusion problem and things seem as expected (i.e., no errors or warnings).

guyer:
Yeah, it's puzzling. I'd think a Laplace problem without a Dirichlet condition to fix the baseline would fail for any mesh and any solver.

PR #721, opened 2020-04-06
Refactor matrices

guyer:
Needed for interprocess communication when writing parallel meshes (#132), but generally beneficial

PR #764, opened 2020-11-07
Disable TVTK tests if its prerequisites aren't met

guyer:


guyer:
Ignore the test failures. They're addressed on #721.

fipy/viewers/vtkViewer/vtkViewer.py
```diff
@@ -15,6 +15,11 @@ def _checkForTVTK():
             from tvtk.api import tvtk
         except ImportError as e:
             from enthought.tvtk.api import tvtk
+
+        # With Python 2.7 end-of-life, bitrot has set in.  While tvtk is
+        # available from conda-forge, it can raise an exception when it
+        # internally does this import.
+        from collections.abc import Sequence
```
wd15, line 22: I don't fully understand why the import is required. If it fails then let it fail and the user gets a warning that it isn't available. How does the extra import impact that?

guyer: If we do that then running CI on Python 2.7 is pointless. We'll never get pull requests, even though we don't really care about TVTK failures and we know why they happen.

wd15: I get it. Tvtk always imports, but fails during the run stage so you're trying to make the failure happen during the import stage.

wd15 (approved):

tkphd (approved):

PR #767, opened 2020-12-01
Introduce SharedTemporaryFile

guyer:
Create a temporary file shared by all MPI ranks


guyer:
I think I broke this by mistake

wd15:
Maybe create a test for this? Also, do you want to add it to the linting step?

PR #773, opened 2021-01-13
Get website documentation to display math

### LINKED ISSUE #762: Equations on Website don't show right
Hi all, 
I got an issue concerning your website, all the equations don't show right for me. Is this a known issue or am I the only one? Do you have any suggestions on how to fix this? I'm using firefox, but I already tried it on different browsers and different systems.
Regards Nico

Example: https://www.ctcms.nist.gov/fipy/examples/diffusion/generated/examples.diffusion.coupled.html
grafik

guyer:
Fixes #762

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,6 +99,16 @@ commands:
 
     steps:
       - run:
+          name: Install LaTeX
+          command: |
+            apt-get --yes update
+            apt-get --yes install texlive-latex-base
+            apt-get --yes install texlive-fonts-recommended
+            apt-get --yes install texlive-fonts-extra
+            apt-get --yes install texlive-latex-extra
+            apt-get --yes install texlive-science
+
+      - run:
           name: Build Docs
           command: |
             source activate ~/project/<< parameters.condaenv >>
@@ -123,6 +133,7 @@ commands:
     steps:
       - run:
           name: Install LaTeX
+          # needed for imgmath
           command: |
             apt-get --yes update
             apt-get --yes install texlive-latex-base
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -498,7 +498,7 @@ jobs:
             apt-get --yes install texlive-fonts-extra
             apt-get --yes install texlive-latex-extra
             apt-get --yes install texlive-science
-            apt-get --yes install texlive-texlive-extra-utils
+            apt-get --yes install texlive-extra-utils
 
       - build_html_docs:
           condaenv: << parameters.condaenv >>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,16 +99,6 @@ commands:
 
     steps:
       - run:
-          name: Install LaTeX
-          command: |
-            apt-get --yes update
-            apt-get --yes install texlive-latex-base
-            apt-get --yes install texlive-fonts-recommended
-            apt-get --yes install texlive-fonts-extra
-            apt-get --yes install texlive-latex-extra
-            apt-get --yes install texlive-science
-
-      - run:
           name: Build Docs
           command: |
             source activate ~/project/<< parameters.condaenv >>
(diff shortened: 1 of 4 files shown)


guyer:
Try as I might, I couldn't get [`sphinx.ext.mathjax`](https://www.sphinx-doc.org/en/master/usage/extensions/math.html#module-sphinx.ext.mathjax) working satisfactorily. The math renders fine, but MathJax is pointedly not TeX, and definitely not LaTeX. We import some packages and define some commands in [conf.py](https://github.com/usnistgov/fipy/blob/master/documentation/conf.py#L220) and I couldn't make enough of it work:
- The `SIunits` package is archaic. There's an [`siunitsx`](https://github.com/burnpanck/MathJax-siunitx) proxy for MathJax 2.x, but it's [buggy](https://tug.org/pipermail/tex4ht/2019q1/002169.html), and it's really clunky to load it in sphinx.
- `changepage` is for the PDF rendering, not web
- defining `\abs` is OK
- Making `\erf` a math operator stymies me. MathJax 3.x has a [`physics`]() package that includes `\erf`, but for the life of me I can't get it to load. MathJax 3.x also broke my def of `\abs`. 

I'm fed up and am giving up. SVG is good enough.

tkphd:
Can we cheat using `\mathrm{erf}`?

guyer:
> Can we cheat using `\mathrm{erf}`?

mayhaps

tkphd (approved):
Changes appear straightforward; docs build locally and look fine (i.e., equations are present & accounted for). MathJax support would be a good follow-up issue.

wd15:
I'm not seeing the equations locally. What am I doing wrong?

wd15:
Works for me. Could we set it to fail if latex doesn't run from some reason?

wd15 (approved):

guyer:
> Could we set it to fail if latex doesn't run from some reason?

I don't know, but it would be desirable. That was why I missed that it didn't build right on CircleCI.

guyer:
> Could we set it to fail if latex doesn't run from some reason?

`imgmath` [logs and raises an error](https://github.com/sphinx-doc/sphinx/blob/7ecf6b88aa5ddaed552527d2ef60f1bd35e98ddc/sphinx/ext/imgmath.py#L169) when it can't find a LaTeX component it needs, but then it [discards the error](https://github.com/sphinx-doc/sphinx/blob/7ecf6b88aa5ddaed552527d2ef60f1bd35e98ddc/sphinx/ext/imgmath.py#L293).

It seems like [the CI command](https://github.com/usnistgov/fipy/blob/b719e006ec0e62daf89ea6c5aec9acad842da870/.circleci/config.yml#L106) needs to grep for "`.* command '.*' cannot be run \(needed for math display\), check the imgmath_.* setting`" and exit non-zero, but doing that, while still outputting the log is beyond my shelling.

Something, something, `tee`?

wd15:
> > Could we set it to fail if latex doesn't run from some reason?
> 
> `imgmath` [logs and raises an error](https://github.com/sphinx-doc/sphinx/blob/7ecf6b88aa5ddaed552527d2ef60f1bd35e98ddc/sphinx/ext/imgmath.py#L169) when it can't find a LaTeX component it needs, but then it [discards the error](https://github.com/sphinx-doc/sphinx/blob/7ecf6b88aa5ddaed552527d2ef60f1bd35e98ddc/sphinx/ext/imgmath.py#L293).
> 
> It seems like [the CI command](https://github.com/usnistgov/fipy/blob/b719e006ec0e62daf89ea6c5aec9acad842da870/.circleci/config.yml#L106) needs to grep for "`.* command '.*' cannot be run \(needed for math display\), check the imgmath_.* setting`" and exit non-zero, but doing that, while still outputting the log is beyond my shelling.
> 
> Something, something, `tee`?

Thanks for looking into it. I would say that it's probably better to get sphinx to return in a failed state rather than having the CI try to interpret the output.

PR #785, opened 2021-03-05
Remove future from setup.py and _setup commands

### LINKED ISSUE #735: pip install fipy fails
Attempting to install FiPy as ``pip install fipy`` fails on clean virtual environments (both Python 3 and 2.7) with an `ImportError` (or `ModuleNotFoundError`).

The reason seems to be that setup.py imports fipy/_setup/build_docs.py, which itself tries to import the `future` package even though it hasn't been installed yet.

Running ``pip install future`` before installing FiPy fixes the problem. Ideally, this workaround shouldn't be necessary and makes it more difficult to have FiPy as a dependency.

Related: #565

guyer:
Getting future from very beginning is problematic.
install_requires will take care of having it when we really need it.

Fixes #735


guyer:
CI failures are not due to this PR

wd15:
Looks good. Definitely a good idea to check pip installation.

wd15 (approved):

PR #790, opened 2021-03-22
fix: stop divide by zero warning in LU solvers

wd15:
Prompted by [this](https://stackoverflow.com/questions/66163197/solving-a-pde-with-1d-zero-gradient-boundary-condition-and-zero-flux-condition-w/66163944#66163944).


wd15:
Test failures seem unrelated to changes

shell.nix
```diff
@@ -31,7 +31,12 @@ in
 
       export OMPI_MCA_plm_rsh_agent=/usr/bin/ssh
 
-      ## To build the docs
+      ## To build the docs (don't use --pure)
```
guyer, line 34: Should this go in #791?

wd15: Done

wd15:
VTK failure on Travis is probably version incompatibility with Numpy.

guyer:
I agree, the VTK errors are safe to ignore

guyer:
I don't understand why the polyxtal example is stalling with SciPy on Py3k

wd15:
> I don't understand why the polyxtal example is stalling with SciPy on Py3k

Is this the first time you've seen that?

guyer:
I can't remember a stall like this for a serial run. I was wondering if the error was somehow negative, so that the `<=` test would fail, but I don't think that's possible and I don't know why that would make it stall.

wd15:
> I can't remember a stall like this for a serial run. I was wondering if the error was somehow negative, so that the `<=` test would fail, but I don't think that's possible and I don't know why that would make it stall.

Not to worry. I'll try and debug it at some point. No rush to finish this.

guyer:
polyxtal is stalling with SciPy on other branches. This doesn't have anything to do with it.

wd15:
@guyer I merged from master and tests are failing for various reasons seemingly not related to the changes.

guyer:
> @guyer I merged from master and tests are failing for various reasons seemingly not related to the changes.

Weird. Looks like a slew of unrelated failures. Should be a joy to debug.

guyer:
Looks like a lot of things started failing [about 5 months ago](https://app.circleci.com/pipelines/github/usnistgov/fipy/326/workflows/b7a16b19-4b8e-4bc8-8734-d51d5aea4ada) when I relaxed the installation constraint on Gmsh from `"gmsh<4.0"` to just `gmsh`. (see #820)

Issue #697, opened 2020-01-24
FiPy cannot partition with Gmsh 4.x


guyer:
FiPy only understands Gmsh's [`msh2` format](http://gmsh.info/doc/texinfo/gmsh.html#MSH-file-format-version-2-_0028Legacy_0029) and [Gmsh 4.x does not export ghost cells correctly in `msh2`](https://gitlab.onelab.info/gmsh/gmsh/issues/733).

- [ ] This constraint should be documented
- [ ] FiPy should support newer Gmsh formats

PR #791, opened 2021-03-22
Add tests for the Nix build

wd15:
Seems to be possible to run completely independent configurations on Travis with the jobs / include syntax.


wd15:
Only seems to be running a single version of the Travis osx tests

wd15:
The following is failing on the Nix build so need to fix that. Also failing locally. Might be to do with the Gmsh version.
```
======================================================================
FAIL: circleQuad (examples.diffusion)
Doctest: examples.diffusion.circleQuad
----------------------------------------------------------------------
  File "/nix/store/z65l1jqvxa58zzwwa3bvglb6asj4y8cv-python3-3.8.5/lib/python3.8/doctest.py", line 2204, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for examples.diffusion.circleQuad
  File "/home/travis/build/usnistgov/fipy/examples/diffusion/circleQuad.py", line 0, in circleQuad
----------------------------------------------------------------------
File "/home/travis/build/usnistgov/fipy/examples/diffusion/circleQuad.py", line 159, in examples.diffusion.circleQuad
Failed example:
    print(phi.allclose(x, atol = 0.035)) # doctest: +GMSH
Expected:
    1
Got:
    False
```

The other failures on Travis aren't to do with these changes, but VTK. Getting
```
numpy.dtype(numpy.character): vtkConstants.VTK_UNSIGNED_CHAR,
    DeprecationWarning: Converting `np.character` to a dtype is deprecated. The current result is `np.dtype(np.str_)` which is not strictly correct. Note that `np.character` is generally deprecated and 'S1' should be used.
```

guyer:
Can you tell if it's a different version of Gmsh or something?

wd15:
> Can you tell if it's a different version of Gmsh or something?

It's giving

    $ gmsh --version      
    4.6.0

Is that supported?

guyer:
> ```
> $ gmsh --version      
> 4.6.0
> ```
> 
> 
> Is that supported?

It looks like it stopped working between Gmsh 4.4 and 4.5. 

I think that failure shouldn't stop you from merging this PR.

wd15:
> It looks like it stopped working between Gmsh 4.4 and 4.5.
> 
> I think that failure shouldn't stop you from merging this PR.

I'll try and fix it. This is an issue that arose from introducing a newer version of Gmsh than Conda's (or whatever FIPy conda recipe is using) version so should be fixed. Is Gmsh fixed in FiPy's conda recipe? At least one of the conda tests is using [3.06](https://travis-ci.com/github/usnistgov/fipy/jobs/493131426#L1629).

guyer:
I think the "fix" is just to relax the tolerance of the test. We definitely should support Gmsh 4.x and, largely, we do.<a href="#Gmsh4" id="Gmsh4ref"><sup>1</sup></a> I've often seen we get sporadic failures when different versions of Gmsh tweak their meshing algorithms. We either need to find a more robust metric to test, or just (continue to) drop the tolerances.

<a id="Gmsh4" href="#Gmsh4ref"><sup>1</sup></a> Remaining issues with Gmsh 4.x
- Some earlier versions of [Gmsh 4.x did not properly record ghost cells in `msh2` meshes](https://gitlab.onelab.info/gmsh/gmsh/-/issues/733), but that's been fixed for awhile.
- We [should support newer MSH formats](https://github.com/usnistgov/fipy/issues/697)
- Our [`conda-forge` feedstock](https://github.com/conda-forge/fipy-feedstock) pulls in a [really old Gmsh 3.0.6](https://travis-ci.com/github/usnistgov/fipy/jobs/493131421#L1669) sometimes
- Our [`conda-forge` feedstock](https://github.com/conda-forge/fipy-feedstock) feedstock doesn't pull in Gmsh at all sometimes (recent Py3k, at least on some platforms (I don't think any of our CIs exhibit this, but they're all still using Python 3.6, which also needs to be updated)).

wd15:
@guyer, @tkphd: please review this when you get a chance. Running Jupyter using Nix on macos should work now.

shell.nix
```diff
@@ -1,37 +1,48 @@
...
-    not_darwin_inputs = pkgs.lib.optionals (! pkgs.stdenv.isDarwin ) [ pythonPackages.jupyter ];
-    not_darwin_pre_shell_hook = if (! pkgs.stdenv.isDarwin) then ''
-      jupyter nbextension install --py widgetsnbextension --user
-      jupyter nbextension enable widgetsnbextension --user --py
-    '' else "";
-    filter_pyamg = builtins.filter (x: ! ((pkgs.lib.hasInfix "pyamg" x.name) && pkgs.stdenv.isDarwin));
+  pkgs = import (builtins.fetchTarball "https://github.com/NixOS/nixpkgs/archive/${tag}.tar.gz") {};
+  pypkgs = pkgs.python3Packages;
+
+  nixes_src = builtins.fetchTarball "https://github.com/wd15/nixes/archive/9a757526887dfd56c6665290b902f93c422fd6b1.zip";
```
tkphd, line 11: What is this magic tarball, and is a user account the proper source for it?

wd15: Good catch. It's a bunch of Jupyter dependencies, [here](https://github.com/wd15/nixes/blob/master/jupyter/default.nix#L42). I don't know why I thought that was a good idea. They can just be listed directly here.

wd15: [TODO]: check if nixpkgs has some sort of built in jupyter-complete expression with all Jupyter add ons.

wd15:
@tkphd should be better now

tkphd (approved):
LGTM!

guyer:
Remaining CI failure is due to [PETSc 3.18](https://github.com/usnistgov/fipy/issues/896) or some CI glitch (for example, why is the [`Test` log](https://dev.azure.com/guyer/FiPy/_build/results?buildId=419&view=logs&j=c5c395d1-7dc3-5472-8d6e-190f0981b61b&t=fd34892f-4c32-5713-43da-f100a9ca44aa) blank, when it can be seen along the way that the tests run successfully to completion?). Regardless, it [fails on `master`](https://dev.azure.com/guyer/FiPy/_build/results?buildId=420&view=logs&j=58fde56f-e4bb-53ea-f269-cde8ff5cac60), too, so no reason to hold up this PR.

PR #793, opened 2021-04-02
Document parallel gmsh

### LINKED ISSUE #609: Parallelizing of Gmsh meshes not clearly documented
I created a 3D spherical mesh in Gmsh 3.0.5 and trying to import to fipy 3.1.3. The following code works when run as `python test_gmsh.py` but crashes when run with `mpirun -np 8 python test_gmsh.py`
```python
#!/usr/bin/env python2
import fipy

mesh = fipy.Gmsh3D('sphere.msh')
```
The error is:
```
  File "test_gmsh.py", line 3, in <module>
    mesh = fipy.Gmsh3D('sphere.msh')
  File "/usr/lib/python2.7/site-packages/fipy/meshes/gmshMesh.py", line 1944, in __init__
    self._orderedCellVertexIDs_data) = self.mshFile.read()
  File "/usr/lib/python2.7/site-packages/fipy/meshes/gmshMesh.py", line 809, in read
    raise GmshException(errStr)
fipy.meshes.gmshMesh.GmshException: Gmsh hasn't produced any cells! Check your Gmsh code.

Gmsh output:
```

If I replace `.msh` file with `.geo` file it works in both cases, however the mesh seems to be different from one generated in Gmsh.

guyer:
* Make clearer how Gmsh meshes are partitioned.
* Document and relax constraints when installing gmsh with conda (constraints were because some versions didn't partition the way FiPy needed, hence including this here).

Fixes #609 
Addresses #697

### DIFF
--- a/INSTALLATION.rst
+++ b/INSTALLATION.rst
@@ -272,7 +272,7 @@ http://www.geuz.org/gmsh/
 
 :term:`Gmsh` is an application that allows the creation of irregular meshes.
 When running in parallel, :term:`FiPy` requires a version of :term:`Gmsh`
->= 2.5 and < 4.0.
+>= 2.5 and < 4.0 or >= 4.5.2.
 
 SciPy
 =====
--- a/INSTALLATION.rst
+++ b/INSTALLATION.rst
@@ -67,13 +67,16 @@ Recommended Method
 * `install Miniconda`_ on your computer
 * run::
 
-    $ conda create --name <MYFIPYENV> --channel conda-forge python=<PYTHONVERSION> fipy
+    $ conda create --name <MYFIPYENV> --channel conda-forge python=<PYTHONVERSION> fipy gmsh
 
   .. note::
 
      This command creates a self-contained conda_ environment and then
      downloads and populates the environment with the prerequisites for
      :term:`FiPy` from the conda-forge_ channel at https://anaconda.org.
+
+     :term:`Gmsh` is an optional package because some versions are
+     incompatible with :term:`FiPy`, so it must be requested explicitly.
 
   .. attention::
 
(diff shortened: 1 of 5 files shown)


guyer:
Failures are unrelated to this PR

INSTALLATION.rst
```diff
@@ -85,7 +88,7 @@ Recommended Method
 
      and for Python 3.x, you should be able to do::
 
-      conda create --name <MYFIPYENV> --channel conda-forge python=3.6 numpy scipy matplotlib pysparse
+      conda create --name <MYFIPYENV> --channel conda-forge "python>3" numpy scipy matplotlib pysparse gmsh
```
wd15, line 91: is `"python>3"` better than `python=3`. I guess that works when python goes to version 4 as well.

guyer: I didn't test. I worried that might give me Python 3.0. We probably *don't* want it to start using Python 4 without warning us.

fipy/meshes/gmshMesh.py
```diff
@@ -1387,6 +1389,11 @@ def _localOverlappingCellIDs(self):
 class Gmsh2D(Mesh2D):
     """Construct a 2D Mesh using Gmsh
 
+    If called in parallel, mesh will be partitioned based on
```
wd15, line 1392: If called in parallel, the mesh will be partitioned based on the value of parallelComm.Nproc

wd15 (approved):

INSTALLATION.rst
wd15: I use `python=3` generally to specify the latest version that conda is using.

wd15: with circle ci or just installing locally

guyer: Thanks

fipy/meshes/gmshMesh.py
guyer: Better, thanks

Issue #796, opened 2021-04-15
examples.phase.polyxtal freezes on CircleCI with Py3k and scipy solvers


guyer:
See [test failure](https://app.circleci.com/pipelines/github/usnistgov/fipy/341/workflows/09b150c2-2aea-4492-9117-5322947ba78c/jobs/4329).

Related to #770?

guyer:
Slowness isolated to `splu()` on CircleCI. scipy/scipy#13999

Every other invocation seems to be fine.

PR #808, opened 2021-05-04
Slow splu

guyer:
Diagnose #796


guyer:
#796 is fixed as a side effect of #822

PR #832, opened 2021-11-23
Add LabelVariable

guyer:
Useful for, e.g., https://pages.nist.gov/pfhub/benchmarks/benchmark8.ipynb

### DIFF
--- a/fipy/variables/labelVariable.py
+++ b/fipy/variables/labelVariable.py
@@ -1,0 +1,74 @@
+from __future__ import unicode_literals
+__docformat__ = 'restructuredtext'
+
+from fipy.variables.cellVariable import CellVariable
+
+__all__ = ["LabelVariable"]
+from future.utils import text_to_native_str
+__all__ = [text_to_native_str(n) for n in __all__]
+
+from scipy import ndimage
+
+class LabelVariable(CellVariable):
+    """Label features in `var` using scipy.ndimage.label
+    
+    Parameters
+    ----------
+    var : ~fipy.variables.cellVariable.CellVariable
+        Field to be labeled.  Any non-zero values in input are counted as
+        features and zero values are considered the background.
+        
+        .. important:
+           Only sensible if `var` is defined on a `...Grid...` Mesh.
+    structure : array_like, optional
+        A structuring element that defines feature connections.
+        `structure` must be centrosymmetric
+        (see ```scipy.ndimage.label`` Notes
+        <https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.label.html#scipy.ndimage.label>`_).
+        If no structuring element is provided,
+        one is automatically generated with a squared connectivity equal to
+        one.  That is, for a 2-D `input` array, the default structuring element
+        is::
+            [[0,1,0],
+             [1,1,1],
+             [0,1,0]]
+    dtype : date-type, optional
+        The desired data-type for the labels. Note that the type must be able
+        to store the largest label, or this Variable will raise an Exception.
+        Default: int.
+    """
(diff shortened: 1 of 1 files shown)


wd15:
A few test cases? It took me a while to figure out what it was supposed to do.

It only works in 2D right? Maybe have an exception in the init with different dimension?

fipy/variables/labelVariable.py
```diff
@@ -0,0 +1,74 @@
...
+        `structure` must be centrosymmetric
+        (see ```scipy.ndimage.label`` Notes
+        <https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.label.html#scipy.ndimage.label>`_).
+        If no structuring element is provided,
+        one is automatically generated with a squared connectivity equal to
+        one.  That is, for a 2-D `input` array, the default structuring element
+        is::
+            [[0,1,0],
+             [1,1,1],
+             [0,1,0]]
```
wd15, line 34: Make this a doctest for example

fipy/variables/labelVariable.py
```diff
@@ -0,0 +1,74 @@
...
+        self.dtype = dtype
+        self._num_features = None
+    
+    def _calcValue(self):
+        """Label features of `var`
+        
+        Side-effect: sets self._num_features
+        """
+        arr = self.var.globalValue.astype(self.dtype)
+        shape = (self.var.mesh.args['nx'], self.var.mesh.args['ny'])
```
wd15, line 60: Only for 2D?

guyer:
> A few test cases? It took me a while to figure out what it was supposed to do.

I'm supposed to have tests just because the code I uploaded didn't work at all?

fipy/variables/labelVariable.py
guyer: Originally only supported 2D, but now supports any dimension that FiPy can grid.

fipy/variables/labelVariable.py
guyer: Done

fipy/variables/cellVariable.py
```diff
@@ -132,23 +132,24 @@ def globalValue(self):
 
     @property
     def ndimage(self):
-        """Global value as an ndarray suitable for scipy.ndimage
+        """Global value as an `ndarray` suitable for `scipy.ndimage`
 
-        The cell values of a `CellVariable` are stored in a 1D ndarray.
+        The cell values of a `CellVariable` are stored in a 1D `ndarray.
```
tkphd, line 137: Need a closing \` on \`ndarray\`

guyer: Fixed

fipy/variables/labelVariable.py
```diff
@@ -0,0 +1,238 @@
...
+        (see ```scipy.ndimage.label`` Notes
+        <https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.label.html#scipy.ndimage.label>`_).
+        If no structuring element is provided,
+        one is automatically generated with a squared connectivity equal to
+        one.  That is, for a 2-D `input` array, the default structuring element
+        is::
+            [[0,1,0],
+             [1,1,1],
+             [0,1,0]]
+    dtype : date-type, optional
```
tkphd, line 188: "data-type"

tkphd (commented):
I wonder what the limits on "distinct non-zero regions" are in [scipy.ndimage.label](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.label.html). Assuming (without direct evidence) that the $\tanh$-like profile of an interface is not considered distinct from the bulk, how close can two bulk particles approach (with adjacent or overlapping interfaces) and remain resolvable?

PR #846, opened 2022-04-01
Get gmsh version from stderr or stdout

guyer:
Since [Gmsh 4.9.0](https://gitlab.onelab.info/gmsh/gmsh/-/commit/a1b7002328b736401c64c8abc3a39d4b1b35a0d2) `gmsh -version` reports to `stdout` instead of `stderr`, which causes FiPy to fail the version check.


wd15:
I get an error on `FAIL: Gmsh2D (fipy.meshes.gmshMesh)`, but it's on both master and this branch so not a regression.

wd15 (approved):

guyer:
> I get an error on `FAIL: Gmsh2D (fipy.meshes.gmshMesh)`, but it's on both master and this branch so not a regression.

What's the error? The CIs are all broken (again!), but none of it is a Gmsh `FAIL` AFAICT. macOS gets a Gmsh `ERROR` because of `dictionary changed size during iteration`.

PR #867, opened 2022-05-20
Document viewer attributes

guyer:


fipy/viewers/matplotlibViewer/matplotlib2DGridContourViewer.py
```diff
@@ -84,25 +101,28 @@ def _plot(self):
...
-        numberOfContours = 10
-        smallNumber = 1e-7
-        zmin = self.norm.vmin
-        zmax = self.norm.vmax
-        diff = zmax - zmin
-
-        if diff < smallNumber:
-            V = numerix.arange(numberOfContours + 1) * smallNumber / numberOfContours + zmin
-        else:
-            V = numerix.arange(numberOfContours + 1) * diff / numberOfContours + zmin
```
wd15, line 96: Is this logic being removed as matplotlib takes care of it now?

fipy/viewers/matplotlibViewer/matplotlib2DContourViewer.py
```diff
@@ -18,10 +18,8 @@ class Matplotlib2DContourViewer(AbstractMatplotlib2DViewer):
     .. _Matplotlib: http://matplotlib.sourceforge.net/
     """
 
-    __doc__ += AbstractMatplotlib2DViewer._test2D(viewer="Matplotlib2DContourViewer")
-
-
-    def __init__(self, vars, title=None, limits={}, cmap=None, colorbar='vertical', axes=None, number=10, levels=None, figaspect='auto', **kwlimits):
+    def __init__(self, vars, title=None, limits={}, cmap=None, colorbar='vertical',
+                 axes=None, number=None, levels=None, figaspect='auto', **kwlimits):
```
wd15, line 22: Line 33 seems to have a stray "float" in front of it (not a change you made, but could be fixed here). Appears in other files as well.

```
limits : dict, optional
    a (deprecated) alternative to limit keyword arguments
float xmin, xmax, ymin, ymax, datamin, datamax : float, optional
    displayed range of data. Any limit set to
    a (default) value of `None` will autoscale.
```

wd15 (approved):

fipy/viewers/matplotlibViewer/matplotlib2DGridContourViewer.py
guyer: Yes. I don't know if it always did, but it certainly does a better job than we do now.

fipy/viewers/matplotlibViewer/matplotlib2DContourViewer.py
guyer: Thanks for that. I've removed everywhere I found it.

PR #876, opened 2022-06-26
Discourage StackOverflow

guyer:
Why, yes, they did piss me off. Why do you ask?


tkphd (approved):
Ha! All righty, let's do this thing.

wd15 (approved):

PR #899, opened 2023-01-12
Manually clean up PETSc objects

### LINKED ISSUE #896: Poor garbage collection with petsc4py 3.18.3 (was "Memory leak in `term.justErrorVector()`", but this isn't strictly a leak)
I'm trying to use a slightly unorthodox solving method that requires me to get the error vector from my set of equations repeatedly. Each time `term.justErrorVector()` is called, it consumes some memory that is never freed. 

Here is some simple code that reproduces the memory increase for me:

```
import fipy
mesh = fipy.Grid1D(nx=100, dx=1)
var1 = fipy.CellVariable(mesh=mesh, hasOld=True, value = 1.0)
eq1 = fipy.DiffusionTerm(var = var1, coeff=1.) == 0

for n in range(1,50000):
    eq1.justErrorVector()
```

This takes a couple of minutes to run and consumes about 5GiB on my system.

I was going to try the other solver implementations (`pysparse` and `scipy`), but `pysparse` won't run on python3 and `scipy` throws an error (which I can submit a bug report for separately.)

guyer:
Garbage collection works differently (worse) in petsc4py 3.18.3 than in 3.17.4.

Fixes #896


tkphd (approved):
LGTM, though I have not tested it to check for memory leaks.

PR #921, opened 2023-04-26
WIP: Propogate boundary conditions to old values

Maxwellfire:
Draft WIP:

Currently fipy does not apply boundary conditions to the old values stored in variables. This means that when you compute some metrics (like the time derivative of a quantity on the boundary), you don't get the expected answer.

I'm making this draft pull request to start a conversation and see if this is something that would be helpful to contribute upstream. I've been using these changes locally for a while.

### DIFF
--- a/fipy/variables/cellVariable.py
+++ b/fipy/variables/cellVariable.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 __docformat__ = 'restructuredtext'
 
 from .meshVariable import _MeshVariable
+from .variable import Variable
 from ..solvers import _MeshMatrix
 from ..tools import numerix
 from ..tools.decorators import deprecate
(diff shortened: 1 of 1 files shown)


guyer:
Thanks for the proposal! This sounds like a reasonable thing to do.

Before merging, I'll need to figure out why the CI has fallen over (and why it hasn't notified me for the last month that it isn't working) and make sure that we get clean tests.

guyer:
@Maxwellfire: I've finally fixed the CI (#925) and your code both passes and seems like a good idea. Would you like to submit this PR for review?

Maxwellfire:
@guyer I think that I should include some code that accounts for deleting boundary conditions as well, and take a final check that my code covers all the cases. Once I do that, I'll submit for review

guyer:
Checking in to see if you've had a chance to polish this PR?

Issue #920, opened 2023-04-26
Should be possible to couple scalar and vector equations


guyer:
As illustrated 

If we attempt to represent

$$
\begin{aligned}
\frac{\partial C}{\partial t} &= \nabla\cdot(D_C \nabla C) + \nabla\cdot(C \vec{v})
\\
\frac{\partial \vec{v}}{\partial t} &= \nabla\cdot(D_\vec{v} \nabla \vec{v})
\end{aligned}
$$

as
```python
import fipy as fp

mesh = fp.Grid2D(nx=10, ny=10)
sca = fp.CellVariable(mesh=mesh, name="sca", hasOld=True)
vec = fp.CellVariable(mesh=mesh, name="vec", rank=1, hasOld=True)

Dsca = 1
Dvec = 2

eq_sca = fp.TransientTerm(var=sca) == fp.DiffusionTerm(coeff=Dsca, var=sca) + fp.ConvectionTerm(coeff=vec, var=sca)
eq_vec = fp.TransientTerm(var=vec) == fp.DiffusionTerm(coeff=[[[Dvec, 0],
                                                               [0, Dvec]]], var=vec)

eq = eq_sca & eq_vec

eq.solve(dt=1)
```
FiPy raises
```python
ValueError: all the input arrays must have same number of dimensions, but the array at index 0 has 1 dimension(s) and the array at index 1 has 2 dimension(s)
```

Solving sequentially is fine:
```python
eq_sca.solve(dt=1)
eq_vec.solve(dt=1)
```

For stability reasons, it may be necessary to use a sequential approach, anyway ([e.g. SIMPLE](https://www.ctcms.nist.gov/fipy/examples/flow/generated/examples.flow.stokesCavity.html)), but FiPy shouldn't prevent trying to solve coupled.

PR #982, opened 2023-12-01
Fix #963: PETSc 3.20.0 broke the world

### LINKED ISSUE #963: PETSc 3.20.0 broke the world
Breakages with PETSc observed in #946 correspond to [PETSc 3.20.0](https://dev.azure.com/guyer/FiPy/_build/results?buildId=844&view=logs&j=6e9fd3b2-3876-5f92-ebcc-0273b412dd37&t=9e30d64e-08f3-56bc-e828-0c4d9a37260a&l=220), (released [2023-09-28](https://petsc.org/release/changes/), conda-forge [petsc4py feedstock updated 2023-10-04](https://github.com/conda-forge/petsc4py-feedstock/commits/main)). Previous [PETSc 3.18.4](https://dev.azure.com/guyer/FiPy/_build/results?buildId=838&view=logs&j=d140db5a-ea74-5114-89c8-047f038a6232&t=e01c99fe-a45f-5995-24fb-d520d2bba4ff&l=213) worked fine.

- [ ] Assembly error
    ```python
    Exception raised:
          File "/Users/guyer/mambaforge-arm/envs/petsc4py319/lib/python3.11/doctest.py", line 1351, in __run
            exec(compile(example.source, filename, "single",
          File "<doctest fipy.matrices.petscMatrix._PETScMatrix.__mul__[10]>", line 1, in <module>
            numerix.allclose(numerix.array((1,2,3),'d') * L1, tmp)
                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
          File "/Users/guyer/Documents/research/FiPy/fipy/fipy/matrices/petscMatrix.py", line 168, in __rmul__
            self.matrix.multTranspose(x, y)
          File "petsc4py/PETSc/Mat.pyx", line 1356, in petsc4py.PETSc.Mat.multTranspose
        petsc4py.PETSc.Error: error code 73
        [0] MatMultTranspose() at /Users/runner/miniforge3/conda-bld/petsc_1695965956423/work/src/mat/interface/matrix.c:2637
        [0] Object is in wrong state
        [0] Not for unassembled vector, did you call VecAssemblyBegin()/VecAssemblyEnd()?
    ```
- [ ] Really slow
    Comparison of the runtimes for FiPy's test suite vs petsc4py version (these are only the 40 slowest tests when using petsc4py 3.20.1):
    Comparison of runtimes vs petsc4py version for 40 slowest test runtimes with petsc4py 3.20.1

    Ratio of runtimes for FiPy's test suite between petsc4py versions
    Ratio of runtimes for FiPy's test suite between petsc4py versions

guyer:
Fixes #963
(Actually, PETSC 3.19.0 broke the world.)

This PR:
- Assembles before `multTranspose` to prevent newly added exception
- Renames `bandwidth` to `nonZerosPerRow` and removes `sizeHint`

  The two were confusingly redundant:
  - PySparse takes `sizeHint`, the number of non-zeros in the matrix.
  - PyTrilinos takes `NumEntriesPerRow`.
  - petsc4py didn't used to be clear what it took, but is now
    documented as number of non-zeros per row (of the local portion
    of the matrix, but we'll ignore that part).
  - scipy doesn't preallocate.
  - Linear algebra
    [defines bandwidth](https://en.wikipedia.org/wiki/Band_matrix#Bandwidth)
    as "number $k$ such that $a_{i,j}=0$ if $|i-j| > k$", which is
    roughly half the number of non-zeros per row (and only applies
    to a band-diagonal matrix).
    Better to be explicit about what we really mean.

  Now all take same parameter and PySparse adjusts as needed.

  `sizeHint` was introduced in @a15d696 (in 2006!) to
  "allow smaller preallocations", but it was never used that way.
  Now, `nonZerosPerRow` can take an array_like to specify row-by-row
  preallocations, which are directly supported by PyTrilinos and petsc4py,
  and can be simulated for PySparse.

  Added `exactNonZeros`, which may have performance benefits for
  PyTrilinos and petsc4py. Currently unused.
- Uses `Term`'s knowledge of own stencil to preallocate more effectively.
  Still doesn't do a good job with vector equations, but that's a deeper change
  (the resolution of which might help #920).

### DIFF
--- a/fipy/meshes/gmshMesh.py
+++ b/fipy/meshes/gmshMesh.py
@@ -2185,10 +2185,13 @@ class Gmsh3D(Mesh):
         ...     os.close(ftmp)
         ... else:
         ...     posFile = None
+        >>> parallelComm.Barrier()
         >>> posFile = parallelComm.bcast(posFile)
+        >>> parallelComm.Barrier()
         >>> f = openPOSFile(posFile, mode='w') # doctest: +GMSH
         >>> f.write(vol) # doctest: +GMSH
         >>> f.close() # doctest: +GMSH
+        >>> parallelComm.Barrier()
 
         >>> f = open(posFile, mode='r') # doctest: +GMSH
         >>> l = f.readlines() # doctest: +GMSH
@@ -2244,6 +2247,7 @@ class Gmsh3D(Mesh):
         >>> print(numerix.allclose(a1, a2)) # doctest: +GMSH
         True
 
+        >>> parallelComm.Barrier()
         >>> if parallelComm.procID == 0:
         ...     os.remove(posFile)
 


guyer:
The one failure is due to an unrelated intermittent race condition

guyer:
Now, the one failure is due to cosmic rays ☢️ 😠

.azure/pipelines.yml
```diff
@@ -258,7 +258,7 @@ stages:
         condition: startsWith(variables.image, 'windows')
 
       - bash: |
-          mamba create --quiet --name wheelEnvironment --channel conda-forge python=3.10 mamba
+          conda create --quiet --name wheelEnvironment --channel conda-forge python=3.10
```
wd15, line 261: Why back to conda?

fipy/matrices/pysparseMatrix.py
```diff
@@ -403,18 +405,24 @@ def __init__(self, rows, cols, bandwidth=0, sizeHint=None, matrix=None, storeZer
...
+            *ignored*
         matrix : ~pysparse.spmatrix.ll_mat
             Pre-assembled Pysparse matrix to use for storage
         storeZeros : bool
             Instructs pysparse to store zero values if possible.
         """
-        sizeHint = sizeHint or max(rows, cols) * bandwidth
         if matrix is None:
             tmpMatrix = spmatrix.ll_mat(1, 1, 1)
+            try:
```
wd15, line 421: Can you do this without a try / except? It just feels wrong here. Does the `asarray` call raise the `TypeError`?

wd15 (approved):

.azure/pipelines.yml
guyer: Because conda now uses libmamba (or at least can be configured to do so, which I do)

fipy/matrices/pysparseMatrix.py
guyer: The `len(nonZerosPerRow)` is what throws the `TypeError`. This is the most effective way to figure out if `nonZerosPerRow` is iterable in Python 2.7. In Py3k, you can (and I do) do an `isinstance(nonZerosPerRow, collections.abc.Iterable)`.

wd15: Oh yeah, FiPy still needs to work with 2.7.

tkphd (approved):
Did not test myself, but the changes look good to me.

PR #985, opened 2023-12-09
Environment pinning

guyer:
Use [`conda-lock`](https://github.com/conda/conda-lock) to pin environments.

This PR:
- tests every commit for breaking changes in the commit
- tests `master` nightly at midnight for breaking changes in dependencies


docs/source/ADMINISTRATA.rst
```diff
@@ -119,6 +119,29 @@ We use the :term:`Azure` and :term:`GitHub Actions` cloud services for
...
 .. _.github/workflows/nix.yml: https://github.com/usnistgov/fipy/blob/master/.github/workflows/nix.yml
 
+===============
+Conda Lockfiles
+===============
+
+The `conda-lock <https://github.com/conda/conda-lock>`_ lockfiles in
+:file:`environments/locks/` can be updated with::
+
+    $ for solver in petsc pysparse scipy trilinos
```
tkphd, line 129: Refactor as a rst code-block.

INSTALLATION.rst
```diff
@@ -61,59 +63,89 @@ Recommended Method
...
-   These days, the most useful channel is `conda-forge`, which provides
+   These days, the most useful channel is conda-forge_, which provides
    everything necessary to install :term:`FiPy`.
 
 * `install Miniconda`_ on your computer
-* run::
+* Install :term:`FiPy` :ref:`REQUIREDPACKAGES` with one of the following:
 
-    $ conda create --name <MYFIPYENV> --channel conda-forge python=<PYTHONVERSION> fipy gmsh
+    * Conda environment files::
```
wd15, line 72: This option will pin all the packages so is the most reproducible, but not particularly upgradable. For most, this is the safest way to generate a FiPy environment that actually works.

wd15: Reverse that. Opposite way around.

INSTALLATION.rst
```diff
@@ -61,59 +63,89 @@ Recommended Method
...
+         with any :file:`environments/<SOLVER>-environment.yml` you wish to
+         use (`conda-merge <https://github.com/amitbeka/conda-merge>`_ may
+         prove useful).  Then execute::
 
-     The `fipy conda-forge`_ package is a convenience. You may choose to
-     install packages explicitly, e.g.,::
+           $ conda env create --name <MYFIPYENV> --file <MYMERGEDENVIRONMENT>.yml
 
-       $ conda create --name <MYFIPYENV> --channel conda-forge python=3 numpy scipy matplotlib-base future packaging mpich mpi4py petsc4py mayavi "gmsh <4.0|>=4.5.2"
+    * `conda-lock <https://github.com/conda/conda-lock>`_ lockfiles::
```
wd15, line 102: This option will pin all the packages so is the most reproducible, but not particularly upgradable. For most, this is the safest way to generate a FiPy environment that actually works.

INSTALLATION.rst
wd15: This option is the most upgradable in the future and probably the best for development.

docs/source/ADMINISTRATA.rst
guyer: It is?

"[A paragraph consisting of two colons ("::") signifies that the following text block(s) comprise a literal block. The literal block must either be indented or quoted (see below).](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#literal-blocks)

wd15 (approved):

wd15:
Maybe point at a blog post such as [this](https://pythonspeed.com/articles/conda-dependency-management/) to help explain the difference between using lock and not using lock.

wd15:
I see, you've already done it.

wd15 (approved):

wd15:
BTW the [Binder link](https://www.ctcms.nist.gov/fipy/INSTALLATION.html#pre-installed-on-binder) never seems to work. Might be best to remove that.

guyer:
> BTW the [Binder link](https://www.ctcms.nist.gov/fipy/INSTALLATION.html#pre-installed-on-binder) never seems to work. Might be best to remove that.

https://mybinder.org [seems to be having an outage right now](https://mybinder.readthedocs.io/en/latest/about/status.html), but even when it's up, it might not be too useful. I've added an #987 to take a look at this.

PR #992, opened 2024-01-10
switch nix recipe to flake

wd15:
Update the Nix recipe as Nix test previously failing.


wd15:
Where is the documentation if the PR is from a fork?

guyer:
> Where is the documentation if the PR is from a fork?

In an [artifact](https://github.com/usnistgov/fipy/actions/runs/7480750633) of the build, but not hosted anywhere

examples/sanitize.cfg
```diff
@@ -1,3 +1,7 @@
-[regex]
+[regex1]
 regex: <Figure size [0-9]{3,4}x[0-9]{3,4}(?:.[0-9])? with [0-9]{1,3} Axes>
 replace: <Figure size>
+
+[regex2]
+regex: <fipy.viewers.matplotlibViewer.matplotlib2DGridViewer.Matplotlib2DGridViewer at .{1,}>
+replace:
```
tkphd, line 7:
```suggestion
[regex2]
regex: <fipy.viewers.matplotlibViewer.matplotlib2DGridViewer.Matplotlib2DGridViewer at .{1,}>
replace: 

```

tkphd (approved):
LGTM

docs/source/NIX-README.rst
```diff
@@ -24,6 +24,7 @@ version of :term:`FiPy`. To test your installation use::
 
    The SciPy solvers are the only availabe solvers currently.
```
tkphd, line 25:
```suggestion
   The SciPy solvers are the only available solvers currently.
```

tkphd (approved):
All tests passed!

flake.nix
```diff
@@ -0,0 +1,49 @@
+{
+  description = "Python environment for fipy";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
```
tkphd, line 5: Maintenance overhead question: @wd15, do you plan to update this URL each May and November?

tkphd (approved):
Still LGTM. Small question about maintenance overhead and hard-coding package versions.

fipy/tools/logging/environment.py
```diff
@@ -55,6 +55,37 @@ def pip_info(python="python"):
...
+         "nix",
+         "derivation",
+         "show",
+         ".#fipy"
+        ], stdout=subprocess.PIPE)
+
+    stdout, _ = p.communicate()
+    stdout = stdout.decode('ascii')
+
+    info["nix_info"] = json.loads(stdout)
```
guyer, line 86: Arguably, this should just return `json.loads(stdout)`, rather than nesting it in a dict. `conda_info()` is a little different because it collects the results of both `conda info` and `conda env export`. One option is that each of these routines should return a dict containing one or more info["???_info"] and then have `fipy.__init__.py` call `_fipy_environment.update(environment.???_info())` on each of them.

flake.nix
wd15: It doesn't need to be updated necessarily, just when we feel like it. It's a single place to update rather than multiple places so in that sense it's good. Can also pin to the latest release so whenever "nix flake update" is executed it gets the latest versions, but that is prone to failure if it's not tested (so still sort of protected from updates even with latest). I'm happy to switch it to latest, but I've been using a specific release for my projects.

fipy/tools/logging/environment.py
wd15: I'll let you decide on that.

PR #995, opened 2024-01-19
Introduce Timer context manager

guyer:
Originally tried to make `Timer()` do logging, too, but I couldn't figure out how to get it to log in the correct scope.
Separating timing from logging substantially simplifies the logic at a modest cost in LoC at point of use.

### DIFF
--- a/fipy/tools/timer.py
+++ b/fipy/tools/timer.py
@@ -1,0 +1,63 @@
+import time
+
+class Timer(object):
+    """Context manager that measures time elapsed in context
+
+    Defaults to nanosecond precision (although probably only microsecond or
+    even millisecond accuracy).
+
+    >>> with Timer() as timer:
+    ...     pass
+    >>> print("elapsed: {elapsed} ns".format(elapsed=timer.elapsed)) # doctest: +ELLIPSIS
+    elapsed: ... ns
+
+    Parameters
+    ----------
+    timer : callable, optional
+        Function that returns a time
+
+            ``timer() -> int or float``
+
+        The difference between successive calls to ``timer()`` should give
+        the elapsed time at the desired resolution and type.
+        (default: :func:`~time.perf_counter_ns`)
+    """
+    def __init__(self, timer=None):
+        self.timer = timer or getattr(time, "perf_counter_ns", self.clock_ns)
+
+        self.start_time = self.stop_time = 0
+        self.running = False
+
+    @staticmethod
+    def clock_ns():
+        """Substitute "nanosecond" timer for Python 2.7
+        """
+        return int(time.clock() * 1e9)
+
+    @property
+    def elapsed(self):
+        """Time measured so far
+        """
+        if self.running:
+            return self.timer() - self.start_time
+        else:
+            return self.stop_time - self.start_time
+
+    def __enter__(self):
+        self.running = True
+        self.start_time = self.stop_time = self.timer()
+
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.stop_time = self.timer()
+        self.running = False
+
+        return False
+
+def _test():
+    import fipy.tests.doctestPlus
+    return fipy.tests.doctestPlus.testmod()
+
+if __name__ == "__main__":
+    _test()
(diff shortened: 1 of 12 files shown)


fipy/tools/timer.py
```diff
@@ -0,0 +1,63 @@
+import time
+
+class Timer(object):
+    """Context manager that measures time elapsed in context
+
+    Defaults to nanosecond precision (although probably only microsecond or
+    even millisecond accuracy).
+
```
tkphd, line 8: `examples.phase.simple` returns elapsed times in $[3\times10^5, 1\times10^8]$ ns.
In the logfile, these large numbers are difficult to read, esp. without thousands separators.
Alongside the accuracy vs. precision argument, I suggest scaling the timer's output from ns to μs or ms, making the unit string part of the Timer class instead of hard-coding, and possibly making the scaling factor and unit user-definable.

tkphd (approved):
The code LGTM. The output looks like (with `input` blocks escaped):

```shell
$ export FIPY_LOG_CONFIG=fipy/tools/logging/serial_config.json
$ python3 simple.py
Solver suite is scipy
{'python': '3.11.5 (main, Sep 11 2023, 13:54:46) [GCC 11.2.0]', 'fipy': '3.4.4+10.g626e154ca', 'numpy': '1.26.3', 'pysparse': 'not installed', 'scipy': '1.11.4', 'matplotlib': '3.8.2', 'mpi4py': 'not installed', 'petsc4py': 'not installed', 'pyamgx': 'not installed', 'PyTrilinos': 'not installed', 'mayavi': 'not installed', 'gmsh': '4.12.0', 'solver': 'scipy'}
False
True
True
True
True
True
True
$ grep -E "[[:digit:]]+[[:space:]]ns" fipy.log
2024-01-19 16:35:15,377 - DEBUG - fipy.terms.binaryTerm._BinaryTerm - _prepareLinearSystem - END _prepareLinearSystem - 7706122 ns
2024-01-19 16:35:15,377 - DEBUG - fipy.solvers.scipy.linearGMRESSolver.LinearGMRESSolver - _solve_ - END precondition - 512 ns
2024-01-19 16:35:15,377 - DEBUG - fipy.solvers.scipy.linearGMRESSolver.LinearGMRESSolver - _solve_ - END solve - 64355 ns
2024-01-19 16:35:15,377 - DEBUG - fipy.terms.binaryTerm._BinaryTerm - solve - END solve - 7903306 ns
2024-01-19 16:35:15,385 - DEBUG - fipy.terms.binaryTerm._BinaryTerm - _prepareLinearSystem - END _prepareLinearSystem - 7594736 ns
2024-01-19 16:35:15,385 - DEBUG - fipy.solvers.scipy.linearLUSolver.LinearLUSolver - _solve_ - END precondition - 33168 ns
2024-01-19 16:35:15,386 - DEBUG - fipy.solvers.scipy.linearLUSolver.LinearLUSolver - _solve_ - END solve - 299729 ns
2024-01-19 16:35:15,386 - DEBUG - fipy.terms.binaryTerm._BinaryTerm - solve - END solve - 8239543 ns
```

fipy/tools/timer.py
guyer: They're not particularly intended to be human-readable. Anything that makes it easier for humans makes it harder for machine parsing. Stupid humans.

The units/type/magnitude of `Timer.elapsed` is a result of the output of the corresponding routine optionally passed to `timer=`. It defaults to [`time.perf_counter_ns()`](https://docs.python.org/3/library/time.html#time.perf_counter_ns). One of the issues with the previous way I was doing this (relying on the log time) was that it was only millisecond precise and had huge uncertainty, rendering useless any stats from small problems and lending quite a bit of uncertainty even to big ones. I was more interested in something unambiguously parsable than I was in something readable.

wd15:
LGTM, I get similar to @tkphd.

I see that the Timer context is being repeated in many different places. Suggests that the context should be in a single place and then what's inside the context should be what is unique to each subclass.

wd15 (approved):

guyer:
> I see that the Timer context is being repeated in many different places. Suggests that the context should be in a single place and then what's inside the context should be what is unique to each subclass.

I see your point. The thing is that what's inside the context is *highly* dependent on the broader context of `._solve_()`. I guess that could be stored as a bunch of `Solver` member variables... I'll think about this in my broader solver refactorization.

Issue #1011, opened 2024-02-15
Error when running coupled.py example


agoncalvescoel:
When I try to run the coupled.py example, I have the following error

```
(MYFIPYENV) anacoelho@HYHNB053:~/fipy$ FIPY_VIEWER=matplotlib python examples/diffusion/coupled.py
QStandardPaths: wrong permissions on runtime directory /run/user/1000/, 0755 instead of 0700
/home/anacoelho/miniconda3/envs/MYFIPYENV/lib/python3.11/site-packages/fipy/viewers/matplotlibViewer/__init__.py:122: UserWarning: Matplotlib1DViewer efficiency is improved by setting the 'datamax' and 'datamin' keys
  return Matplotlib1DViewer(vars=vars, title=title, axes=axes, **kwlimits)
Uncaught exception:
Traceback (most recent call last):
  File "/home/anacoelho/fipy/examples/diffusion/coupled.py", line 125, in <module>
    exec(fipy.tests.doctestPlus._getScript())
  File "<string>", line 103, in <module>
  File "/home/anacoelho/miniconda3/envs/MYFIPYENV/lib/python3.11/site-packages/fipy/viewers/__init__.py", line 131, in Viewer
    raise ImportError("Failed to import a viewer: %s" % str(errors))
ImportError: Failed to import a viewer: ['matplotlib: can only concatenate str (not "int") to str']
```

guyer:
Thank you for reporting. This is a duplicate of #961. I'll prioritize pushing out a fix.

PR #1013, opened 2024-02-16
Convert integer to string in operator variable represenation

### LINKED ISSUE #961: Representation of index variables is broken
```python
>>> import fipy as fp
>>> mesh = fp.Grid2D(nx=3, ny=4)
>>> mesh.cellCenters[0]
  File ~/mambaforge-arm/envs/fipy311/lib/python3.11/site-packages/IPython/core/formatters.py:699 in __call__
    return repr(obj)
  File ~/Documents/research/FiPy/fipy/fipy/variables/operatorVariable.py:307 in __repr__
    return self._getRepresentation()
  File ~/Documents/research/FiPy/fipy/fipy/variables/operatorVariable.py:87 in _getRepresentation
    representation = self._py3kInstructions(op=self.op, style=style, argDict=argDict, id=id, freshen=freshen)
  File ~/Documents/research/FiPy/fipy/fipy/variables/operatorVariable.py:211 in _py3kInstructions
    stack.append(stack.pop(-2) + "[" + stack.pop() + "]")
TypeError: can only concatenate str (not "int") to str
```

The `stack.pop()` needs to be wrapped in `str()`, or the whole thing should use f-string formatting.

guyer:
Fixes #961
Addresses #1011

### DIFF
--- a/fipy/variables/operatorVariable.py
+++ b/fipy/variables/operatorVariable.py
@@ -373,8 +373,8 @@ def _testBinOp(self):
 
     Test that the representation of indexing works correctly
 
-        >>> ttns(v1[1]._getRepresentation())
-        'Variable(value=array([1, 2, 3, 4]))[1]'
+        >>> ttns(v1[1]._getRepresentation()) # doctest: +ELLIPSIS
+        'Variable(value=array([1, 2, 3, 4]))[...]'
 
     Check that unit works for a `binOp`
 
--- a/fipy/variables/operatorVariable.py
+++ b/fipy/variables/operatorVariable.py
@@ -208,7 +208,7 @@ def _OperatorVariableClass(baseClass=object):
                     # Removed in Python 3
                     stack.append("`" + stack.pop() + "`")
                 elif ins.opname == 'BINARY_SUBSCR':
-                    stack.append(stack.pop(-2) + "[" + stack.pop() + "]")
+                    stack.append(stack.pop(-2) + "[" + str(stack.pop()) + "]")
                 elif ins.opname == 'RETURN_VALUE':
                     s = stack.pop()
                     if style == 'C':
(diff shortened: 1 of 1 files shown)


tkphd:
I'm not pythonic enough to understand what this syntax change means. It looks like you're changing the code from taking the second element of the array, to I have no idea what the `[...]`-th element means.

guyer:
> I'm not pythonic enough to understand what this syntax change means. It looks like you're changing the code from taking the second element of the array, to I have no idea what the `[...]`-th element means.

The test is asking for the represenation of the lazy evaluation of the second element in the array. Under Python 2.7, it returned `Variable(value=array([1, 2, 3, 4]))[index]`. (After this fix) under Python 3, it returns `Variable(value=array([1, 2, 3, 4]))[1]`. `# doctest: +ELLIPSIS` enables me to write a test that accepts either output.

The fact that humans won't know what that means is OK; this test is under `_testBinOp()` because it's not intended as documentation.

tkphd:
> > I'm not pythonic enough to understand what this syntax change means. It looks like you're changing the code from taking the second element of the array, to I have no idea what the `[...]`-th element means.
> 
> The test is asking for the represenation of the lazy evaluation of the second element in the array. Under Python 2.7, it returned `Variable(value=array([1, 2, 3, 4]))[index]`. (After this fix) under Python 3, it returns `Variable(value=array([1, 2, 3, 4]))[1]`. `# doctest: +ELLIPSIS` enables me to write a test that accepts either output.
> 
> The fact that humans won't know what that means is OK; this test is under `_testBinOp()` because it's not intended as documentation.

Ah-ha! Cool, thank you.

guyer:
nix is the devil

tkphd:
> nix is the devil

and not spoken of in polite company

PR #1015, opened 2024-02-16
Fix explanation for no-flux steady-state failure

### LINKED ISSUE #1010: examples.diffusion.mesh1D No-flux - steady-state doesn't always give zero
When I run `mesh1D.py` , the other parts of the script run fine, but the part on the figure below gives me different results
image

The blue line does not appear at all
image

And when it uns the validation of numerix, gives me False
image

guyer:
Fixes #1010

### DIFF
--- a/examples/diffusion/mesh1D.py
+++ b/examples/diffusion/mesh1D.py
@@ -720,24 +720,30 @@ If we reset the initial condition
 
 and solve the steady-state problem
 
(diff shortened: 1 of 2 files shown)


wd15:
Works for me.

wd15:
If you merge #992 then we won't have the Nix failures at least

guyer:
> If you merge #992 then we won't have the Nix failures at least

Oh, so I suppose it's all my fault, then?

PR #1030, opened 2024-04-02
[WIP] Harmonize solvers

guyer:


guyer:
Silenced Codacy warnings about "Method XYX has NN lines of code (limit is 50)". These methods do *not* have more than 50 lines of code. They have more than 50 lines of docstring. Shut up Codacy.

guyer:
Documentation still needs performance plots

fipy/solvers/__init__.py
```diff
@@ -179,18 +178,45 @@ def _import_mesh_matrices(suite):
...
+INDEX_TYPE = _MeshMatrix.INDEX_TYPE
+
 from fipy.tests.doctestPlus import register_skipper
 
 register_skipper(flag='PYSPARSE_SOLVER',
                  test=lambda: solver_suite == 'pysparse',
                  why="the Pysparse solvers are not being used.",
                  skipWarning=True)
 
+register_skipper(flag='PETSC_SOLVER',
```
wd15, line 190: Obviously, a lot of these repeated patterns could be condensed.

wd15:
A few things I noticed that aren't important:

- The caption below [here](https://pages.nist.gov/fipy/en/harmonize_solvers/USAGE.html#solver-suites) seems to be all the way over to the left
- Also, [here](https://pages.nist.gov/fipy/en/harmonize_solvers/USAGE.html#solving-in-parallel) PyTrilinos only has support up to Python 3.7 (not for Python 3 generically, right?)
- "Hopefully, we will learn how to optimize..." [here](https://pages.nist.gov/fipy/en/harmonize_solvers/USAGE.html#running-under-python-2), you have already done this and the next release will have these improvements (or doesn't this release have those improvements?)

wd15 (approved):

guyer:
> A few things I noticed that aren't important:
> 
> * The caption below [here](https://pages.nist.gov/fipy/en/harmonize_solvers/USAGE.html#solver-suites) seems to be all the way over to the left

Yep

> * Also, [here](https://pages.nist.gov/fipy/en/harmonize_solvers/USAGE.html#solving-in-parallel) PyTrilinos only has support up to Python 3.7 (not for Python 3 generically, right?)

AFAIK, PyTrilinos is not intrinsically limited to 3.7 (3.8 on macOS), but the conda builds aren't compatible with anything newer.

> * "Hopefully, we will learn how to optimize..." [here](https://pages.nist.gov/fipy/en/harmonize_solvers/USAGE.html#running-under-python-2), you have already done this and the next release will have these improvements (or doesn't this release have those improvements?)

Good catch, thanks

guyer:
> Documentation still needs performance plots

This is still true

wd15:
The table for convergence is great.

guyer:
> Overall, I think this is ready to go. The ideas below can be placed in an issue and implemented after the release.

Thanks very much for your comments

> Ideas:
> 
> * A lot of the information on this page could be summed up in a table of solver v feature with ticks and crosses. This might be a lot less verbose and have a lot less of the warning boxes.

Good idea.

> * Add a label to the x-axis of part (a) of the first figure.

Matplotlib does this with shared axes (there's no y-axis label in (b), either). I've tried unsharing and then it seems very cluttered with the same label over and over (see the prepare fraction plot).

> * The performance comparison section could do with a little explanation about the history of FiPy and what you just did. Give the take away in a few sentences and why we need this analysis.

Good point.

> * The second figure is a little confusing. It's not clear why the PCG solver in in bold. Is this because it's the baseline?

The intent was to show the range of possible behaviors with the fine lines and to compare one particular (well-performing) example with the thick lines.

> * Maybe add Figure numbers and refer to them in the text.

I'd love to. Can't figure out how in Sphinx.

> * There is quite a lot of variation in the prepare time / elapsed time results within each suite. Maybe say why different solvers are better / worse in this regard in the same suite?

Some solvers suck at solving?

> * It says " For this problem, Trilinos has the lowest ratio of prepare to elapsed time". All suites have some solvers that do well in this regard. I'm not sure you can claim that Trilinos is better overall. Scipy seems to get the lowest for the majority of solvers. Linear PCG does better with Trilinos.

I was thinking about unpreconditioned LinearPCGSolver when I wrote this. The fact that SciPy has many lines that are very low is exactly the point I was trying to make. All SciPy solves of a particular mesh size have the same prepare time. If prepare/elapsed is small, it's because the solver is slow, not because the prepare is fast.

> * How are the prepare time / elapsed time > 1 for petsc?

Not > 1; asymptotically 1. Prepare time is dreadful for PETSc >= 3.19 and FiPy 3.4.4.

> * Having two keys in a plot is hard work for the reader. Maybe make two plots for the parallel performance plot each with a single key

Let them work!

>   * I'm thinking that having the curves for the parallel laws is confusing. Maybe split those out into separate plots.

Separate plots for Amdahl and Gunther fits makes sense.

guyer:
The [nix issue](https://github.com/usnistgov/fipy/actions/runs/17302151101/job/49115854638?pr=1030) doesn't seem to have anything to do with this PR.
- It fails the same way on my laptop
- It fails with `master`
- numpy is old (1.26.4)
- nix is frozen
- FiPy hasn't touched anything related

Based on googling, maybe the GitHub Action image has changed from `x86_64` to `aarch64-darwin`?

wd15:
Congrats!

PR #1052, opened 2024-06-21
Test nix with the Issue1049-numpy_2.0.0_breaks_things branch

wd15:


guyer:
@wd15: Having tried a bunch of diagnostics and getting nothing to show up in the nix tests, I think what's happening is that:
- your nix installation already has FiPy in it?
- `nix develop --command bash -c "python setup.py test 2> test.out" || true` runs the fipy tests in the source directory, but when the tests `import fipy`, they don't import from the source directory, they import the `fipy` that's already installed.

This would explain why we see changes in the doctests (because it's extracting those from the source code), but we don't see changes to the code (because it's running the code that's already installed).

My naive attempt to do something about that (2dd8321) results in 
```
× This environment is externally managed
╰─> This command has been disabled as it tries to modify the immutable
    `/nix/store` filesystem.
    
    To use Python with Nix and nixpkgs, have a look at the online documentation:
    <https://nixos.org/manual/nixpkgs/stable/#python>.
```
I don't (want to) understand nix well enough to work around that.

wd15:
It is definitely using the updated version of `fipy/tests/doctestPlus.py` as it prints on at the top of the file and prints to the locally cloned version. Installing with `pip install` won't work as Nix doesn't allow this for obvious reasons. Although it's using the cloned version of `fipy/tests/doctestPlus.py`, it is never failing with the `raise Exception` in the `testmod` function. What is going on?

wd15:
Ok, although `fipy/tests/doctestPlus.py` is imported, the debugging version of `fipy/tests/test.py` is never imported. Why is that?

wd15:
Other than when fipy uses `fipy.test()`, `fipy.tests.test` is never imported in the fipy source. It's handed over to setup.py which knows what to do.

guyer:
Let me do some experiments

wd15:
Oh! Building the egg info actually made it use the modified test module.

guyer:
> Oh! Building the egg info actually made it use the modified test module.

Yes. Turns out it isn't a nix thing, per se. Fresh checkout, without building egg_info, doesn't execute the relevant tweaks to `printoptions`. After building egg_info, for any reason, all is OK.

Still not right for this particular Action; it doesn't appear to be running the tests after it builds egg_info. I'll leave that to you.

wd15:
> > Oh! Building the egg info actually made it use the modified test module.
> 
> Yes. Turns out it isn't a nix thing, per se. Fresh checkout, without building egg_info, doesn't execute the relevant tweaks to `printoptions`. After building egg_info, for any reason, all is OK.
> 
> Still not right for this particular Action; it doesn't appear to be running the tests after it builds egg_info. I'll leave that to you.

Very good. Will get this cleaned up.

wd15:
Just one failure now, which also occurs locally as well. Some sort of reordering error with Scipy / Numpy interaction.

PR #1067, opened 2024-08-03
Prevent setuptools 72.x

guyer:
Work on broken environment for nightly CI vs PR CI


guyer:
Spelling failing because [cgit.freedesktop.org is down](https://downforeveryoneorjustme.com/cgit.freedesktop.org) (source of en_US.dic and en_US.aff)

Issue #1079, opened 2024-11-01
Intermittent failures with shared parallel files


guyer:
CI tests are failing intermittently because of possible race conditions with shared files in parallel:
- [`SharedTemporaryFile`](https://dev.azure.com/guyer/FiPy/_build/results?buildId=1429&view=logs&j=27df24f5-e7d5-549e-9e2f-0469bcd2e5d0&t=e3e7200e-73a6-593c-bd03-712459ca1714&l=3135)
- [`GmshFile`](https://dev.azure.com/guyer/FiPy/_build/results?buildId=1427&view=logs&j=6e9fd3b2-3876-5f92-ebcc-0273b412dd37&t=9399fa50-d076-5fc8-faf5-3d956ac1eea6&l=3288)

Issue #1072, opened 2024-09-17
skfmm does not seem to be listed as a dependency when running the tests


wd15:
See [here](https://dev.azure.com/guyer/a5dd49f0-3688-4211-8a43-0bfb5a6c1a9e/_apis/build/builds/1362/logs/339)

```
2024-09-17T00:18:08.0531699Z FAILED (failures=9)
2024-09-17T00:18:08.0532121Z !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
2024-09-17T00:18:08.0532683Z Skipped 83 doctest examples because the `tvtk` package cannot be imported
2024-09-17T00:18:08.0533311Z !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
2024-09-17T00:18:08.0533766Z 
2024-09-17T00:18:08.0534712Z python      3.12.5 | packaged by conda-forge | (main, Aug  8 2024, 18:24:51) [MSC v.1940 64 bit (AMD64)]
2024-09-17T00:18:08.0535326Z fipy        3.4.5+8.gd8d9f88fe
2024-09-17T00:18:08.0535670Z numpy       2.1.1
2024-09-17T00:18:08.0535999Z pysparse    not installed
2024-09-17T00:18:08.0536332Z scipy       1.14.1
2024-09-17T00:18:08.0536841Z matplotlib  3.9.2
2024-09-17T00:18:08.0537182Z mpi4py      not installed
2024-09-17T00:18:08.0537526Z petsc4py    not installed
2024-09-17T00:18:08.0537881Z pyamgx      not installed
2024-09-17T00:18:08.0538227Z PyTrilinos  not installed
2024-09-17T00:18:08.0538612Z mayavi      4.8.2
2024-09-17T00:18:08.0538898Z gmsh        4.12.2
2024-09-17T00:18:08.0539235Z solver      scipy
2024-09-17T00:18:08.0539471Z 
2024-09-17T00:18:08.5110119Z 
2024-09-17T00:18:08.5114240Z ##[debug]
```

guyer:
[...]

PR #1082, opened 2024-11-19
Handle exporting anonymous conda environment

guyer:
`conda export env --name` doesn't work when active_prefix_name is a path (environment doesn't have a name)


wd15 (approved):

guyer:
Known failures (#1072 and #1079) are unrelated to this PR

Issue #1002, opened 2024-02-05
Setting nx, ny, nz to non-positive integer should generate an error


guyer:
`nx`, `ny`, and `nz` are cardinal numbers of grid cells. They should not be allowed to be floats or non-positive integers (zero probably needs to be allowed for parallel partitioning).

```python
>>> mesh = fp.Grid1D(nx=0.01, dx=10)
>>> mesh
UniformGrid1D(dx=10, nx=0)
```

PR #1083, opened 2024-11-20
issue1002-Setting nx, ny, nz to non-positive integer should generate an error

sridhar-mani:
Tried to correct the issue No.1002.Included validation checks for the inputs.

Setting nx, ny, nz to non-positive integer should generate an error #1002


guyer:
Thank you very much for the contribution. As you can see, this PR currently generates a lot of test failures. Most seem to be because of where you do the check. [For non-uniform grids, the default value for `nx`, `ny`, and `nz` is `None`](https://github.com/usnistgov/fipy/blob/e7afd1a58a6d9a852a11655145288f8cddb87ba7/fipy/meshes/nonUniformGrid2D.py#L23), which fails your checks.

- [ ] I think the correct place to make this check is in [`_NonuniformNumPts.calcNs()`](https://github.com/usnistgov/fipy/blob/e7afd1a58a6d9a852a11655145288f8cddb87ba7/fipy/meshes/builders/utilityClasses.py#L90) and [`_UniformNumPts.calcNs()`](https://github.com/usnistgov/fipy/blob/e7afd1a58a6d9a852a11655145288f8cddb87ba7/fipy/meshes/builders/utilityClasses.py#L90).
- [ ] I don't see any need for a `try ... except ...` block. The check should just fail.
    - If you really don't want the failure, then it should raise a `Warning`, but definitely not print out an error message.
- [ ] Please remove the addition of [`.vscode/settings.json`](https://github.com/usnistgov/fipy/pull/1083/files#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357). It doesn't seem to have anything to do with this PR. If you'd like to separately propose VSCode settings, please make a separate PR and explain your rationale (I don't use VSCode, so I don't know what makes sense as a project setting vs. a personal setting).

sridhar-mani:
have made the changes there

guyer:
Test failures are unrelated to this PR

PR #1105, opened 2025-04-06
Generate root .rst files with sphinx_readme (#1104)

### LINKED ISSUE #1104: README files viewed on GitHub.com render poorly
## Observed behavior

Several of the main documentation files in the root directory of the project do not render well when viewed via github.com, especially the README.rst which is the first thing people see when visiting the project.  That may leave a bad first impression on visitors to the GitHub repository, impeding adoption of the software and inhibiting potential contributors.

These files have Sphinx extensions in them that do not play well with the RST renderer GitHub uses:

* CHANGELOG.rst
* INSTALLATION.rst
* README.rst

The README.rst file shows numerous sections that should be hidden, and has a broken TOC:
Image

The INSTALLATION.rst has sections that should be hidden, and broken terms:
Image

The CHANGELOG shows broken terms:
Image

Curiously, these `.rst` are fine when viewed in GitHub

* DISCLAIMER.rst
* LICENSE.rst

## Expected behavior

The files in the root directory, which are by convention the most fundamental documentation for a project, should render OK when viewed in GitHub.

## Proposed fix

Alter the documentation production pipeline, which depends on Sphinx, to generate these files in a way that is compatible with GitHub's rendering engine.  Consider using [Sphinx README](https://github.com/TDKorn/sphinx-readme) to fix up the files that don't render well. Alternatively, explore generating Markdown alternatives for the root directory with something like [Sphinx Markdown Builder](https://pypi.org/project/sphinx-markdown-builder/)

## Supporting information

I believe GitHub uses https://github.com/github/markup to determine what toolchain to use to render files. Then it uses an unspecified internal toolchain to produce the rendered HTML. For `rst` files `github/markup` it requires `docutils` to be installed. The `docutils` project, if you use it to render HTML from `.rst`, produces output that is similar to the badly rendered HTML that GitHub shows. GitHub _definitely_ does not have a fully Sphinx-compatible rendering pipeline for Restructured Text documents.

GitHub does a decent job of rendering simpler Restructured Text files and the output of  [Sphinx README](https://github.com/TDKorn/sphinx-readme) doesn't have the same problems.

obscurerichard:
* Use https://sphinx-readme.readthedocs.io/en/latest/index.html to generate .rst files in root directory
* Original files are all now in docs/sources/ directory

Resolves #1104

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,17 @@
+.. |.examples.diffusion.mesh1D| replace:: ``examples.diffusion.mesh1D``
+.. _.examples.diffusion.mesh1D: https://github.com/usnistgov/fypy/blob/2730380faa50d1b2f9cc6d5a1b5d9f9f4a41bd1f/examples/diffusion/mesh1D.py
+.. |.faq| replace:: Frequently Asked Questions
+.. _.faq: https://pages.nist.gov/fipy/en/latest/FAQ.html#faq
+.. |.FiPy| replace:: FiPy
+.. _.FiPy: https://pages.nist.gov/fipy/en/latest/glossary.html#term-FiPy
+.. |.installation| replace:: Installation
+.. _.installation: https://pages.nist.gov/fipy/en/latest/INSTALLATION.html#installation
+.. |.Python| replace:: Python
+.. _.Python: https://pages.nist.gov/fipy/en/latest/glossary.html#term-Python
+.. |.usage| replace:: Using FiPy
+.. _.usage: https://pages.nist.gov/fipy/en/latest/USAGE.html#usage
+
+
 ========
 Overview
 ========
@@ -8,27 +22,9 @@ Overview
 | |GitHub|_ |PyPI|_  |CondaForge|_ |Binder|_
 | |gitter|_ |OpenHub|_
 
-.. only:: latex
 
-   :term:`FiPy` is an object oriented, partial differential equation (PDE)
-   solver, written in :term:`Python`, based on a standard finite volume
-   (FV) approach. The framework has been developed in the Materials Science
-   and Engineering Division (MSED_) and Center for Theoretical and
-   Computational Materials Science (CTCMS_), in the Material Measurement
-   Laboratory (MML_) at the National Institute of Standards and Technology
-   (NIST_).
 
-   The solution of coupled sets of PDEs is ubiquitous to the numerical
-   simulation of science problems.  Numerous PDE solvers exist, using a
-   variety of languages and numerical approaches. Many are proprietary,
-   expensive and difficult to customize.  As a result, scientists spend
-   considerable resources repeatedly developing limited tools for
-   specific problems.  Our approach, combining the FV method and :term:`Python`,
-   provides a tool that is extensible, powerful and freely available. A
-   significant advantage to :term:`Python` is the existing suite of tools for
-   array calculations, sparse matrices and data rendering.
-
-The :term:`FiPy` framework includes terms for transient diffusion,
+The |.FiPy|_ framework includes terms for transient diffusion,
 convection and standard sources, enabling the solution of arbitrary
 combinations of coupled elliptic, hyperbolic and parabolic PDEs. Currently
 implemented models include phase field :cite:`BoettingerReview:2002`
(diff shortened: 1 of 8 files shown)


docs/source/README.rst
```diff
@@ -1 +1,172 @@
...
+Overview
+========
+
+
+
+| |Tests|_ |Documentation|_ |nix|_
+| |GitHub|_ |PyPI|_  |CondaForge|_ |Binder|_
+| |gitter|_ |OpenHub|_
+
+.. only:: latex
```
guyer, line 11: I need to think about how to handle these `only:: latex` blocks. We do this because we have a custom [index.html](https://github.com/usnistgov/fipy/blob/master/docs/source/_templates/index.html). The README gets linked from the "About FiPy" link on the home page. At the time I did all this, I didn't care for the redundant text between those two pages, so excluded it in the HTML with `only:: latex`. The thing is, that text *should* show up when viewing the README in GitHub or `cat README.rst` from the command-line.

The right answer is probably to tolerate the redundancy.

README.rst
```diff
@@ -1,3 +1,17 @@
+.. |.examples.diffusion.mesh1D| replace:: ``examples.diffusion.mesh1D``
+.. _.examples.diffusion.mesh1D: https://github.com/usnistgov/fypy/blob/2730380faa50d1b2f9cc6d5a1b5d9f9f4a41bd1f/examples/diffusion/mesh1D.py
+.. |.faq| replace:: Frequently Asked Questions
+.. _.faq: https://pages.nist.gov/fipy/en/latest/FAQ.html#faq
+.. |.FiPy| replace:: FiPy
+.. _.FiPy: https://pages.nist.gov/fipy/en/latest/glossary.html#term-FiPy
```
guyer, line 6: Links like `.FiPy` generate [spelling errors](https://dev.azure.com/guyer/FiPy/_build/results?buildId=1607&view=logs&j=590c7ebb-8554-5341-5372-44ea3649a043&t=27e67363-6dd9-5c72-e3dc-8f391dff9547&l=65).

They can be added to the dictionary, but maybe there's a sphinx_readme option that avoids the error?

guyer (commented):
Thanks very much for following up on what you suggested at OSCON!

guyer:
> * Possible TODO: add a note about these 3 files being generated somewhere else in the documentation.

This should probably go in [ADMINISTRATA.rst](https://github.com/obscurerichard/fipy/blob/fix/README.rst-github-rendering/docs/source/ADMINISTRATA.rst)

docs/source/conf.py
```diff
@@ -351,3 +352,22 @@ def setup(app):
     target = target.strip()
     nitpick_ignore.append((dtype, target))
 
+
+# sphinx-readme options
+# See https://sphinx-readme.readthedocs.io/en/latest/configuration/configuring.html
+html_context = {
+    'display_github': True,
+    'github_user': 'usnistgov',
+    'github_repo': 'fypy',
```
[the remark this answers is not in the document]
obscurerichard: Ah, good catch. Fixed in the latest pushes.

README.rst
obscurerichard: I poked at the [sphinx_readme docs](https://sphinx-readme.readthedocs.io/en/latest/configuration/configuring.html) and sources but couldn't find an option that treats `:term:` or glossary items specially. This is the first time I've really used Restructured Text and Sphinx so I'm pretty cold on it.  It looks like there's a bunch of rules in  [.pyspelling.yml](https://github.com/usnistgov/fipy/blob/master/.pyspelling.yml) that already deal with restructured text - maybe adding a rule there to deal with this class of problems is in order instead.

docs/source/README.rst
obscurerichard: Yeah, this seemed a bit awkward and fiddly, just having the docs be the same makes sense to _me_. When in doubt, simplify, right?

guyer: Agreed. FYI, I'm out for a bit, but will work on this part when I get back later next week.

README.rst
guyer: > It looks like there's a bunch of rules in [.pyspelling.yml](https://github.com/usnistgov/fipy/blob/master/.pyspelling.yml) that already deal with restructured text - maybe adding a rule there to deal with this class of problems is in order instead of hard-coding more terms in [.en-custom.txt](https://github.com/usnistgov/fipy/blob/master/.en-custom.txt).

`.pyspelling.yml` is definitely the way to go. Let me see if I can get the rule right.

guyer: I have this working, although pyspelling's filters are weird and completely undocumented

guyer:
> This should probably go in [ADMINISTRATA.rst](https://github.com/obscurerichard/fipy/blob/fix/README.rst-github-rendering/docs/source/ADMINISTRATA.rst)

Done

docs/source/README.rst
guyer: No more `.. only:: latex`

obscurerichard:
@guyer thank you for working with me on this! I'm glad I could help.

docs/source/ADMINISTRATA.rst
```diff
@@ -155,6 +155,29 @@ The `conda-lock <https://github.com/conda/conda-lock>`_ lockfiles in
...
+
+ * Make edits to the corresponding files in :file:`docs/source/`.
+ * Run::
+
+    $ make -C docs html
+
+   to re-render the 3 affected files.
+
+ * Add and commit the resulting changes.
+ bu
```
obscurerichard, line 180: ooops, looks like there is a straggler line here

guyer: Good catch

guyer:
This is raising errors for our latexpdf build: (TDKorn/sphinx-readme#12) (turns out it always was and I misinterpreted the CI failure as something else). If I can't figure out how to disable sphinx-readme for latexpdf, I'm going to have to roll back this change.

PR #1157, opened 2025-09-15
remove python build step as not necessary

wd15:


guyer:
Entrypoints need to be established: [ImportError: No viewers found. Run `python -m build` or similar.](https://github.com/usnistgov/fipy/actions/runs/17741406208/job/50416360173?pr=1157#step:9:52)

PR #1179, opened 2026-02-12
Fix BibTeX citation: escape ampersand in journal name

### LINKED ISSUE #1141: Add \ before & so that bibtex citation is correct.
When copying the citation for a TeX document, the `&` in the journal name does not have a `\` before it and thus causes issues when compiling a TeX document if you just copy the bibtex text as a citation. I'd suggest changing it to
```
 journal =   {Computing in Science \& Engineering}, 
```

https://github.com/usnistgov/fipy/blob/2b1086a2747fa0b66e553106900f312d62212ff4/docs/source/_static/fipy.bib#L6

crabby-rathbun:
Fixes #1141

The ampersand in the journal name "Computing in Science & Engineering" needs to be escaped as  for proper BibTeX formatting when users copy the citation directly into TeX documents.

This is a simple 1-character fix that prevents compilation errors.

### DIFF
--- a/docs/source/_static/fipy.bib
+++ b/docs/source/_static/fipy.bib
@@ -3,7 +3,7 @@
    title =     {{FiPy}: Partial Differential Equations with {P}ython},
    publisher = {IEEE},
    year =      2009,
-   journal =   {Computing in Science & Engineering},
+   journal =   {Computing in Science \& Engineering},
    volume =    11,
    number =    3,
    pages =     {6-15},


comradef191:
The author of this PR is an AISlop spambot account.
Do not merge; Advise blocking author & closing PR.

See https://github.com/matplotlib/matplotlib/pull/31132 for further details wherein the spambot gets mad at the matplotlib authors for refusing to merge its PR there and generates an aggressive blogpost in response.

PR #1182, opened 2026-02-24
Access VTK without saving to disk

isaac-blanc:
As best I can tell, the only way to export `CellVariable` objects to VTK is by using `VTKViewer.plot()`, which saves the result to a file. This little PR adds the `VTKViewer.raw()` method, which returns a VTK object.

This is useful if you want to view visualisations on the fly, without saving to the disk, but you want more flexibility than is offered by the existing integration between FiPy and Mayavi. In my `demo.py` script, I show how this can be used with [pyvista](https://docs.pyvista.org/).

My method shares a lot of duplicated code with `VTKViewer.plot()`. I'm happy to neaten this up if @guyer wants this pulled into main. This PR is just a first draft.

### DIFF
--- a/fipy/viewers/vtkViewer/vtkViewer.py
+++ b/fipy/viewers/vtkViewer/vtkViewer.py
@@ -109,8 +109,8 @@ class VTKViewer(AbstractViewer):
 
     def raw(self):
         """
-        Returns a VTK object, which can be passed to a visualiser of choice. In
-        the example below, we pass this to the popular visualisation library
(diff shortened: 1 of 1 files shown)


isaac-blanc:
Sorry for the slow response! Have just refactored `.plot()` and `.raw()` as requested. I also improved the docstring a touch. Let me know if any more needs done, or if you have any feedback.

guyer:
Thanks for the contribution @isaac-blanc!

isaac-blanc:
A pleasure!

PR #1200, opened 2026-05-28
Fix: ResidualTerm can be negated and scaled (#885)

### LINKED ISSUE #885: Negation of ResidualTerm throws error
Background: I am trying to implement newton iterations, as described in [here](https://gist.github.com/guyer/f29c759fd7f0f01363b8483c7bc644cb). In this document, they use a `fipy.ResidualTerm` to avoid having to recalculate the entire simple equation as an explicit source term

The issue:
Residual terms don't seem to implement coefficients, and so are not able to be negated. A simple piece of code demonstrates the problem: 
```
import fipy
-fipy.ResidualTerm(equation=None)
```
which throws the error:
```

  File "/tmp/ipykernel_36715/862288887.py", line 1, in <cell line: 1>
    -fipy.ResidualTerm(equation=None)

  File "/home/max/Tools/anaconda3/envs/fipy/lib/python3.9/site-packages/fipy/terms/nonDiffusionTerm.py", line 23, in __neg__
    return self.__class__(coeff=-self.coeff, var=self.var)

TypeError: __init__() got an unexpected keyword argument 'coeff'
```

Looking at the code, this seems to happen because the signature of the Residual term is different from every other term. Every other term is fully defined by a coefficient and a variable, but `ResidualTerm` is defined by an equation.

gaoflow:
# Fix: `ResidualTerm` can be negated and scaled

Fixes #885.

## Problem

`ResidualTerm` inherits `__neg__` and `__mul__` from `_NonDiffusionTerm`, which assume a `(coeff, var)` constructor signature:

```python
def __neg__(self):
    return self.__class__(coeff=-self.coeff, var=self.var)
```

`ResidualTerm`'s constructor is `__init__(self, equation, underRelaxation=1.)` — there is no `coeff` keyword — so unary minus *and* scalar multiplication both raise

```text
TypeError: __init__() got an unexpected keyword argument 'coeff'
```

That breaks the obvious cases (`-RT(eq)`, `2.0 * RT(eq)`) but, more insidiously, it also breaks every equation that puts a `ResidualTerm` on the right-hand side of `==`, because `Term.__sub__` distributes negation through the term tree:

```text
lhs == diffusion + ResidualTerm(eq)
# Term.__eq__   →  self - other
# Term.__sub__  →  self + (-other)
# _BinaryTerm.__neg__ →  (-self.term) + (-self.other)
# → _NonDiffusionTerm.__neg__(ResidualTerm) → TypeError
```

So the workaround documented in [#885](https://github.com/usnistgov/fipy/issues/885#issuecomment-1268172069) (parenthesise the LHS so the RHS never gets negated) is the only way `ResidualTerm` works inside an equation today.

## Fix

Override `__neg__` and `__mul__` on `ResidualTerm` to fold the sign/scale into a small internal `_scale` multiplier that `_buildMatrix` applies to the residual vector at solve time:

```python
self.coeff = CellVariable(mesh=var.mesh,
                          value=vec * self.underRelaxation * self._scale)
```

`underRelaxation` keeps its documented semantics (a relaxation factor); the signed/scaled contribution is the product of `underRelaxation` and `_scale`. `__neg__` and `__mul__` construct a fresh `ResidualTerm` and stamp `_scale` on it, so chains like `-(-RT)` and `2 * (3 * RT)` compose correctly.

## Verification

End-to-end at the RHS-vector level:

```python
>>> mesh = Grid1D(nx=10, dx=1.)
>>> v = CellVariable(mesh=mesh, hasOld=True, value=np.arange(10, dtype=float))
>>> v.updateOld()
>>> eq = TransientTerm(var=v) == DiffusionTerm(var=v, coeff=1.)
>>> rt = ResidualTerm(equation=eq)
>>> sandbox = CellVariable(mesh=mesh, hasOld=True, value=0.); sandbox.updateOld()
>>> _, _, b_pos = rt._buildAndAddMatrices(sandbox, _ScipyMeshMatrix, dt=0.1)
>>> _, _, b_neg = (-rt)._buildAndAddMatrices(sandbox, _ScipyMeshMatrix, dt=0.1)
>>> _, _, b_3   = (3. * rt)._buildAndAddMatrices(sandbox, _ScipyMeshMatrix, dt=0.1)
>>> np.allclose(np.asarray(b_neg), -np.asarray(b_pos))
True
>>> np.allclose(np.asarray(b_3),  3 * np.asarray(b_pos))
True
```

`b_pos = [1, 0, ..., 0, -1]` (a diffusion-only residual on a linear ramp), `b_neg = [-1, 0, ..., 0, 1]`, `b_3 = [3, 0, ..., 0, -3]`. The negation/scaling is applied exactly once per build.

All four originally-broken paths from the issue now work:

```python
-fipy.ResidualTerm(equation=eq)                                  # was TypeError
2.0 * fipy.ResidualTerm(equation=eq)                             # was TypeError
lhs == rhs + fipy.ResidualTerm(equation=eq)                      # was TypeError
lhs == rhs - fipy.ResidualTerm(equation=eq)                      # was TypeError
```

and the doctest-discovered division `RT / 4.` routes through `__rmul__` so it works too.

Doctests live in `fipy/terms/residualTerm.py` and exercise both the `_scale` bookkeeping and the numerical RHS contribution. The module is added to `fipy/terms/test.py` so `fipy_test` picks them up:

```console
$ python -c "import doctest, fipy.terms.residualTerm as m; \
              print(doctest.testmod(m, verbose=False))"
TestResults(failed=0, attempted=29)

$ python -c "import unittest, fipy.terms.test as t; \
54 13
```

The 54 vs the previous 50 includes the new `residualTerm` doctests. The 13 failures are all pre-existing numpy-2.x repr-formatting drift in other terms (`[[ 1.  ...]]` vs `[[1. ...]]`), unrelated to this patch.

## Scope

`fipy/terms/residualTerm.py` (the fix + doctests) and `fipy/terms/test.py` (one line to register the new doctest module). No public API change — `ResidualTerm(equation, underRelaxation=1.)` keeps its signature; the new `_scale` is an internal field that defaults to 1 and is preserved by negation/scaling only.

### DIFF
--- a/fipy/terms/residualTerm.py
+++ b/fipy/terms/residualTerm.py
@@ -27,7 +27,7 @@ class ResidualTerm(_ExplicitSourceTerm):
        >>> eq = TransientTerm(var=v) == DiffusionTerm(var=v, coeff=1.)
        >>> rt = ResidualTerm(equation=eq)
 
-    valid for use with subtraction or scalar pre-multiplication.  Negation
+    valid for use with subtraction or scalar multiplication.  Negation
     flips the sign of the residual contribution:
 
        >>> sandbox = CellVariable(mesh=mesh, hasOld=True, value=0.)
--- a/fipy/terms/residualTerm.py
+++ b/fipy/terms/residualTerm.py
@@ -1,6 +1,7 @@
 __docformat__ = 'restructuredtext'
 
 from fipy.terms.explicitSourceTerm import _ExplicitSourceTerm
+from fipy.terms import TermMultiplyError
 from fipy.variables.cellVariable import CellVariable
 
 __all__ = ["ResidualTerm"]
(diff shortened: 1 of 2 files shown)


guyer:

PR #1202, opened 2026-05-28
Fix: UniformGrid* now define _cellToFaceDistanceVectors and _cellDistanceVectors (#706)

### LINKED ISSUE #706: UniformGrids don't define _cellToFaceDistanceVectors or _cellDistanceVectors
As [reported on StackOverflow](https://stackoverflow.com/q/60073399/2019542), the `UniformGrid` classes do not define `_cellToFaceDistanceVectors` or `_cellDistanceVectors`, which are needed for upwinding and for [Robin conditions](https://www.ctcms.nist.gov/fipy/documentation/USAGE.html#applying-robin-boundary-conditions).

gaoflow:
# Fix: `UniformGrid*` now define `_cellToFaceDistanceVectors` and `_cellDistanceVectors`

Fixes #706.

## Problem

`NonUniformGrid*` meshes compute `_cellToFaceDistanceVectors` and `_cellDistanceVectors` in `Mesh.__init__` (via `_calcFaceToCellDistAndVec` and `_calcCellDistAndVec`). `UniformGrid*` meshes inherit from `AbstractMesh` (not `Mesh`) and **never** computed them — so on any `Grid1D` / `Grid2D` / `Grid3D`:

```python
>>> import fipy
>>> mesh = fipy.Grid1D(nx=4, dx=1.)
>>> mesh.cellToFaceDistanceVectors
AttributeError: 'UniformGrid1D' object has no attribute '_cellToFaceDistanceVectors'
>>> mesh.cellDistanceVectors
AttributeError: 'UniformGrid1D' object has no attribute '_cellDistanceVectors'
```

This blocks upwinding and Robin boundary conditions on uniform meshes — exactly what is documented in [USAGE.html#applying-robin-boundary-conditions](https://www.ctcms.nist.gov/fipy/documentation/USAGE.html#applying-robin-boundary-conditions) and reported on [StackOverflow](https://stackoverflow.com/q/60073399/2019542). Linked from the original issue body.

## Fix

Add the two properties on the `UniformGrid` base class. Each subclass already defines `_faceCenters`, `_cellCenters`, and `faceCellIDs` with the same conventions as the non-uniform meshes (shape `(dim, ncells)`, `(dim, nfaces)`, `(2, nfaces)` with the second row of `faceCellIDs` masked along the boundary), so the calculation is the same as the existing reference implementation in `Mesh._calcFaceToCellDistAndVec` / `Mesh._calcCellDistAndVec`:

```python
@property
def _cellToFaceDistanceVectors(self):
    tmp = MA.repeat(self._faceCenters[..., numerix.NewAxis, :], 2, 1)
    tmp = tmp - numerix.take(self._cellCenters, self.faceCellIDs, axis=1)
    return tmp

@property
def _cellDistanceVectors(self):
    tmp = numerix.take(self._cellCenters, self.faceCellIDs, axis=1)
    tmp = tmp[..., 1, :] - tmp[..., 0, :]
    tmp = MA.filled(MA.where(MA.getmaskarray(tmp),
                             self._cellToFaceDistanceVectors[:, 0],
                             tmp))
    return tmp
```

## Verification

For 1D, 2D and 3D, the new uniform properties produce values **bit-equal** to the non-uniform reference at the same `dx, dy, dz, nx, ny, nz`:

```python
>>> from fipy.meshes.nonUniformGrid2D import NonUniformGrid2D
>>> u  = fipy.Grid2D(nx=3, ny=2, dx=1., dy=2.)
>>> nu = NonUniformGrid2D(nx=3, ny=2, dx=1., dy=2.)
>>> import numpy as np
>>> np.allclose(np.array(u._cellToFaceDistanceVectors),
...             np.array(nu._cellToFaceDistanceVectors))
True
>>> np.allclose(np.array(u._cellDistanceVectors),
...             np.array(nu._cellDistanceVectors))
True
>>> u._cellToFaceDistanceVectors.shape, u._cellDistanceVectors.shape
((2, 2, 17), (2, 17))
```

Doctests covering both the value match and the shape contract are added in each `UniformGrid*D._test` (already in the `fipy.meshes.test` suite).

## Test impact

```console
$ python -c "import unittest, fipy.meshes.test as t; \
# baseline (master @ d8c21fd26): 81 19
# with this patch:                81 19
```

Identical to the master baseline. The 19 pre-existing failures are all numpy-2.x repr drift in `cylindrical/spherical*` meshes, unrelated to this patch.

## Scope

Four files: `fipy/meshes/uniformGrid.py` (the two new properties on the base class) plus a small `_test` doctest in each of `uniformGrid1D.py`, `uniformGrid2D.py`, `uniformGrid3D.py` to lock in the value match. No public API change — `cellToFaceDistanceVectors` and `cellDistanceVectors` were already exposed on `AbstractMesh` via `property(lambda s: s._cellToFaceDistanceVectors)`; this patch just makes them resolve instead of raising.

### DIFF
--- a/fipy/meshes/uniformGrid.py
+++ b/fipy/meshes/uniformGrid.py
@@ -1,6 +1,8 @@
 __docformat__ = 'restructuredtext'
 
 from fipy.meshes.abstractMesh import AbstractMesh
+from fipy.tools import numerix
+from fipy.tools.numerix import MA
 
 __all__ = ["UniformGrid"]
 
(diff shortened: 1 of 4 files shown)


guyer:

PR #1206, opened 2026-05-29
fix(solvers): pyamgx solvers crash on instantiation due to precon attribute typo

### LINKED ISSUE #1199: object has no attribute 'precon' in `pyamgx` solvers
Hi,

As a minimum working example, I copied the following from the [original example](https://github.com/shwina/pyamgx/blob/main/examples/FiPy/diffusion.ipynb) of using the amgx solver and run with `FIPY_SOLVERS=pyamgx python3 fipy_trial.py`.

```
from fipy.solvers.pyamgx.pyAMGXSolver import PyAMGXSolver

cfg_dict = {
	"config_version": 2,
	"determinism_flag": 1,
	"exception_handling": 1,
	"solver": {
	    "print_grid_stats": 1,
	    "store_res_history": 1,
	    "solver": "GMRES",
	    "print_solve_stats": 1,
	    "obtain_timings": 1,
	    "preconditioner": {
		"interpolator": "D2",
		"print_grid_stats": 1,
		"solver": "AMG",
		"smoother": "JACOBI_L1",
		"presweeps": 2,
		"selector": "PMIS",
		"coarsest_sweeps": 2,
		"coarse_solver": "NOSOLVER",
		"max_iters": 1,
		"interp_max_elements": 4,
		"min_coarse_rows": 2,
		"scope": "amg_solver",
		"max_levels": 24,
		"cycle": "V",
		"postsweeps": 2,
	    },
	    "max_iters": 100,
	    "monitor_residual": 1,
	    "gmres_n_restart": 10,
	    "convergence": "RELATIVE_INI_CORE",
	    "tolerance": 1e-06,
	    "norm": "L2",
	},
}

solver = PyAMGXSolver(cfg_dict)
```

I get the following stack trace Even though there is clearly a `solver` key in `cfg_dict`:

```
AMGX version 2.5.0
Built on May 26 2026, 10:05:58
Compiled with CUDA Runtime 12.8, using CUDA driver 12.8
The AMGX_initialize_plugins API call is deprecated and can be safely removed.
  File "[filename]", line 39, in <module>
    solver = PyAMGXSolver(cfg_dict)
             ^^^^^^^^^^^^^^^^^^^^^^
  File "/path/.conda/envs/pfc/lib/python3.11/site-packages/fipy/solvers/pyamgx/pyAMGXSolver.py", line 43, in __init__
    self.config_dict["solver"]["max_iters"] = self.iterations
    ~~~~~~~~~~~~~~~~^^^^^^^^^^
KeyError: 'solver'
The AMGX_finalize_plugins API call is deprecated and can be safely removed.
```

I have also tried using other [pyAMGXsolvers](https://github.com/usnistgov/fipy/tree/master/fipy/solvers/pyamgx) but any attempt to instantiate them runs into an issue where none of the solvers have the `precon` attribute, even if explicitly passed. I can provide more details for this if there is not a simple solution to the first issue.

Thanks for any help.

gaoflow:
Fixes #1199.

### Problem
Instantiating any pyamgx-backed solver raises immediately, before any GPU work:

```python
>>> from fipy.solvers.pyamgx import LinearGMRESSolver
>>> LinearGMRESSolver()
AttributeError: 'LinearGMRESSolver' object has no attribute 'precon'
```

`PyAMGXSolver.__init__` checks `self.precon`, but the base `Solver` stores the
preconditioner on `self.preconditioner` (set from the `precon` argument via
`value_or_default(precon, default_preconditioner)`). There is no `self.precon`
attribute, so construction fails.

This matches the maintainer’s diagnosis in #1199 ("it looks like there’s a typo … that should be `if self.preconditioner is not None: self.preconditioner._applyToSolver(...)`"), and the reporter confirmed that renaming it removes the error.

### Fix
1. Use `self.preconditioner` instead of the non-existent `self.precon`.
2. Forward `precon=precon` to the base `Solver.__init__`, matching `scipySolver`
   and `petscKrylovSolver`. Without this, both an explicitly passed `precon`
   **and** the pyamgx solvers’ class `DEFAULT_PRECONDITIONER` were silently
   dropped — the base initializes `self.preconditioner` from its `precon` arg,
   which previously defaulted to `"default"` → `None` for every pyamgx solver.

```python
super(PyAMGXSolver, self).__init__(tolerance=tolerance, criterion=criterion,
                                   iterations=iterations, precon=precon)
...
if self.preconditioner is not None:
    self.preconditioner._applyToSolver(self.config_dict["solver"])
```

### Verification
- `py_compile` clean; no remaining `self.precon` reference.
- Verified by code inspection and consistency with the sibling `scipySolver` /
  `petscKrylovSolver` `__init__` signatures, plus the maintainer diagnosis and
  reporter confirmation in #1199.
- Not runtime-tested locally: pyamgx requires the AMGX library and an NVIDIA
  GPU, which are unavailable in this environment, so I could not add a
  CI-runnable regression test without that dependency. Happy to add one if there
  is a pyamgx-capable CI lane I can target.

### DIFF
--- a/fipy/solvers/pyamgx/pyAMGXSolver.py
+++ b/fipy/solvers/pyamgx/pyAMGXSolver.py
@@ -35,15 +35,16 @@ class PyAMGXSolver(Solver):
         **kwargs
             Other AMGX solver options
         """
-        super(PyAMGXSolver, self).__init__(tolerance=tolerance, criterion=criterion, iterations=iterations)
+        super(PyAMGXSolver, self).__init__(tolerance=tolerance, criterion=criterion,
+                                           iterations=iterations, precon=precon)
 
         # update solver config:
         self.config_dict = self.CONFIG_DICT.copy()
 
         self.config_dict["solver"]["max_iters"] = self.iterations
 
-        if self.precon is not None:
-            self.precon._applyToSolver(self.config_dict["solver"])
+        if self.preconditioner is not None:
+            self.preconditioner._applyToSolver(self.config_dict["solver"])
 
         smoother = self.value_or_default(smoother, self.default_smoother)
         if smoother is not None:


guyer:

PR #1210, opened 2026-06-04
Honor iterations in SciPy LinearLUSolver instead of capping at 10

### LINKED ISSUE #1208: SciPy LinearLUSolver has artificial limit of 10 iterations
Since iterations were introduced, the SciPy `LinearLUSolver` has been [constrained to a maximum of 10 iterations](https://github.com/usnistgov/fipy/blob/d8c21fd266f7792bdc9b3221b683e76b28e2a220/fipy/solvers/scipy/linearLUSolver.py#L68). None of the other `LinearLUSolver` implementations do this.

gaoflow:
Fixes #1208.

## Problem

The SciPy `LinearLUSolver` runs its iterative-refinement loop over `range(min(self.iterations, 10))` ([`linearLUSolver.py:68`](https://github.com/usnistgov/fipy/blob/d8c21fd266f7792bdc9b3221b683e76b28e2a220/fipy/solvers/scipy/linearLUSolver.py#L68)), silently capping the user-requested iteration count at 10. The PETSc and Trilinos `LinearLUSolver` implementations both loop over `range(self.iterations)` with no such cap, so the SciPy backend was inconsistent and ignored `iterations` set above 10.

As discussed in the issue, LU refinement rarely needs more than a couple of sweeps, but users should still be able to set `iterations` to whatever they want.

## Fix

Drop the cap so the SciPy solver honors `iterations` like the other backends. The refinement loop still `break`s as soon as the residual meets the tolerance, so well-conditioned systems converge in the same one-to-few sweeps as before; only an explicit request for more (or fewer) sweeps is now respected.

## Test

Adds a regression doctest to `LinearLUSolver`, registered for the SciPy solver suite in `fipy/solvers/test.py` (mirroring the backend-conditional registration in `fipy/matrices/test.py`). With an unreachable `tolerance=0.` the loop runs the full requested count, so `iterations=12` records 12 refinement sweeps — it recorded 10 before this change.

### DIFF
--- a/fipy/solvers/scipy/linearLUSolver.py
+++ b/fipy/solvers/scipy/linearLUSolver.py
@@ -10,6 +10,28 @@ __all__ = ["LinearLUSolver"]
 
 class LinearLUSolver(ScipySolver):
     """Interface to :term:`LU`-factorization in :ref:`SciPy`.
+
+    The number of iterative refinement sweeps honors the `iterations`
+    argument, consistent with the other `LinearLUSolver` implementations
+    (`GH #1208 <https://github.com/usnistgov/fipy/issues/1208>`_).  Forcing
+    an unreachable tolerance runs the full count rather than capping at 10:
+
+    >>> import warnings
+    >>> import fipy as fp
+    >>> mesh = fp.Grid1D(nx=10)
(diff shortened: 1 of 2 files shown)


guyer:

PR #1224, opened 2026-08-23
Preserve lazy dependencies in cached convection stencil

shin4141:
## Summary

- Preserve the lazy dependency from mutable diffusion coefficients into the cached Peclet calculation.
- Keep large-Peclet sign changes and boundary constraint weights connected to the same lazy graph.
- Add doctest regression coverage for common coefficient scaling and zero-diffusion transitions.

Fixes #1223.

## Verification

- Focused `abstractConvectionTerm` doctest: 97 attempted, 0 failed.
- `fipy.terms.test._suite --scipy`: 50 passed.
- `fipy.variables.test._suite --scipy`: 112 passed.
- Bounded mutation matrix: common/different coefficient changes, zero and negative diffusion, unit-bearing coefficients, six representative convection term paths, Grid1D/Grid2D, and cache/dependency retention passed.
- Full modules + examples with SciPy: 564 of 565 passed. The sole `fipy.solvers.solver.Solver._test` failure is unchanged in an exact pristine master worktree (`1 failed / 32 attempted` in both candidate and pristine).
- `git diff --check`: passed.

## Environment limitations

Local verification used Python 3.12.13, NumPy 2.3.5, SciPy 1.18.1, and the serial SciPy solver on macOS arm64. PETSc, Trilinos, PyAMGX, parallel execution, Windows, and Linux were not locally verified and are left to upstream CI and maintainer review.

## AI assistance disclosure

AI assistance disclosure: I used the OpenAI Codex desktop coding agent in a human-owned local workflow to inspect FiPy source and history, analyze issue #1223, draft the candidate code and doctest text, and run and organize the reproduction, dependency-graph, unit, cache, performance, negative-control, and regression checks. The production changes and the new doctest portions in `fipy/terms/abstractConvectionTerm.py` were AI-assisted and initially generated or drafted by Codex. I reviewed the complete proposed change, its purpose, the three dependency-preserving production changes, the regression coverage, and the stated environment limitations, and I take responsibility for the submission. I will personally review maintainer feedback and remain responsible for any response or follow-up change; no agent will interact with maintainers without my oversight.

PR #1225, opened 2026-08-26
Issue1223 convection-diffusion stencil is not invalidated when mutable coefficients change

### LINKED ISSUE #1223: Convection–diffusion stencil is not invalidated when mutable coefficients change
When a DiffusionTerm and either ConvectionTerm or ExponentialConvectionTerm appear in the same equation, changing their FiPy-variable coefficients after equation construction does not produce the same result as rebuilding the equation with the updated coefficients.
Both coefficient expressions report their updated values correctly. The terms also update correctly when tested independently. The inconsistency only appears when diffusion and convection are combined, suggesting that the cached Peclet-dependent convection–diffusion stencil is not invalidated when its coefficient dependencies change.

**Versions tested**

The issue reproduces with:
FiPy 3.4.1
FiPy 4.0.1

**Reproducer**

See the attached code:

[Minimal_FiPy_ExponentialConvectionTerm_Stencil_Cache_Reproducer.txt](https://github.com/user-attachments/files/30904646/Minimal_FiPy_ExponentialConvectionTerm_Stencil_Cache_Reproducer.txt)

The reproducer contains four tests:
DiffusionTerm alone
ExponentialConvectionTerm alone
DiffusionTerm + ExponentialConvectionTerm
DiffusionTerm + ConvectionTerm
In the combined tests, a mutable scale multiplies both the diffusion and convection coefficients:
```
scale = CellVariable(mesh=mesh, value=1.0)

diffusion_coefficient = scale.faceValue
convection_coefficient = (scale * energy).faceGrad

equation = (
DiffusionTerm(coeff=diffusion_coefficient, var=solution)
- ExponentialConvectionTerm(
coeff=convection_coefficient,
var=solution,
)
== 0.0
)
```
After solving once, the common scale is changed:
```
scale.setValue(2.0)
```
Multiplying both terms by the same positive constant should multiply the complete steady-state equation by that constant and therefore leave its solution unchanged.
Both coefficient expressions visibly double, but reusing the existing equation changes the solution. Rebuilding an identical equation at the new scale restores the original solution.

**Actual output**

With FiPy 4.0.1:
```
A. DiffusionTerm only
reuse versus rebuild error: 0.0
error from expected 1/2 scaling: 0.0

B. ExponentialConvectionTerm only
reuse versus rebuild error (+1 -> +2): 0.0
reuse versus rebuild error (+1 -> -1): 0.0

C. DiffusionTerm + ExponentialConvectionTerm
diffusion expression doubled: True
convection expression doubled: True
invariance error when equation is reused: 0.05436365208454913
invariance error when equation is rebuilt: 0.0
reused versus rebuilt error: 0.05436365208454913

D. DiffusionTerm + regular ConvectionTerm
diffusion expression doubled: True
convection expression doubled: True
invariance error when equation is reused: 0.05060033348695059
invariance error when equation is rebuilt: 0.0
reused versus rebuilt error: 0.05060033348695059
```
FiPy 3.4.1 produces the same behavior.

**Expected behavior**

After ```scale.setValue(2.0)```, reusing the existing combined equation should produce the same solution as:
the solution at ```scale = 1```, because the entire steady-state equation was multiplied by the same constant; and
a newly constructed equation using the updated coefficients.

**Actual behavior**

The coefficient variables themselves update, but the reused combined equation produces a different solution. Rebuilding the equation produces the expected solution.

**Suspected cause**

This appears to involve cached convection weights or a cached Peclet-dependent stencil. The diffusion and convection terms behave correctly independently, but their combined discretization appears to retain information derived from the coefficients used during the first solve.
Is there a supported way to invalidate or refresh this stencil after coefficient variables change, or should this happen automatically through FiPy’s normal variable-dependency mechanism?

**Practical impact**

This affects coupled multiphysics calculations where diffusion or drift coefficients change during a simulation—for example, temperature-dependent drift–diffusion systems. Currently, the reliable workaround is to reconstruct the continuity equations whenever the temperature-dependent coefficients change.

**Disclaimer**

AI was used to draft the simplest possible reproduction after I discovered the issue on a larger problem.

guyer:
Fixes #1223


shin4141:
Thanks for preparing #1225. I ran one additional focused countercase against the current exact head ad86bfafad610f7317300c6a499ec0b89acfe828.

The proportional-scaling case is fixed, and Peclet / alpha now refresh correctly, but a zero-diffusion transition still leaves the cached boundary matrix and RHS stale when the existing equation is reused.

On the exact current head:

reused vs rebuilt solution: 1.1809969335461346
Peclet:                    0.0
alpha:                      0.0
matrix:                     0.5414940825367983
RHS:                        0.5414940825367982
alpha in constraintL deps:  false
alpha in constraintB deps:  true

The remaining dependency break appears to be here:

alpha_constraint = numerix.where(
    var.faceGrad.constraintMask,
    1.0,
    alpha,
)

At first assembly, this removes alpha from the cached constraintL dependency graph, while constraintB still retains a direct dependency on alpha.

This is the same zero-diffusion / boundary-constraint countercase I had tested during the earlier investigation in #1224. In that investigation, making this boundary expression lazy as well caused reused and rebuilt solution / Peclet / alpha / matrix / RHS to all agree.

I used OpenAI Codex to run this focused exact-head countercheck and organize the results. I reviewed the reproducer, the current #1225 diff, and the outputs above, and I am posting this myself.

fipy/terms/abstractConvectionTerm.py
```diff
@@ -122,20 +123,56 @@ def _getWeight(self, var, transientGeomCoeff=None, diffusionGeomCoeff=None):
...
+            diagonalSign = self._getDiagonalSign(transientGeomCoeff,
+                                                  diffusionGeomCoeff)
+            diagonalSign = 1 - 2 * numerix.all(diagonalSign < 0)
+            pecletLarge = 1e+20 * geometrySign * diagonalSign
 
             if diffusionGeomCoeff is None or diffusionGeomCoeff[0] is None:
                 peclet = pecletLarge
             else:
-                diffCoeff = diffusionGeomCoeff[0].numericValue
+                diffCoeff = diffusionGeomCoeff[0]
```
wd15, line 175: Was this a bug or doesn't it matter?

wd15:
LGTM assuming @shin4141 comment is addressed.

wd15 (approved):

guyer:
> The proportional-scaling case is fixed, and Peclet / alpha now refresh correctly, but a zero-diffusion transition still leaves the cached boundary matrix and RHS stale when the existing equation is reused.

This has nothing to do with changing diffusion to zero. `alpha_constraint`  is set to the initial value of `alpha` and sees no changes in diffusion. Changing `alpha_constraint` to be a variable expression in terms of `alpha` is the correct solution.

PR #1227, opened 2026-09-18
Fix typo in table of contents for section numerical

StefRe:
The filename is [discret.rst](https://github.com/usnistgov/fipy/blob/master/docs/source/numerical/discret.rst) instead of discrete.rst. This typo prevents discret.rst from being included in the documentation:  

grafik  
  
  
  
Changing the reference is less instrusive than renaming the file.


guyer:
Thanks for fixing this error introduced in d8c21fd

